PR #4, opened 2020-02-10
Apply the "let it crash" strategy to plug-in failures 

gousiosg:
Our current error handling code delegates too much error handling responsibility to the plug-ins. In this branch, I try to apply the [Let it crash](https://wiki.c2.com/?LetItCrash) design strategy. The idea is that the plug-ins should not do ANY error handling and always raise unchecked exceptions (subclasses of RuntimeException). Methods in plug-ins should not throw / declare checked exceptions, except if those are handled locally in the plug-in. The plug-in runner must install a generic exception handler for RuntimeExceptions and perform the necessary corrective actions.

This is work in progress; PR submitted just to check current status.

### DIFF
--- a/analyzer/dummy-analyzer/src/main/java/eu/fasten/analyzer/dummyanalyzer/DummyAnalyzerPlugin.java
+++ b/analyzer/dummy-analyzer/src/main/java/eu/fasten/analyzer/dummyanalyzer/DummyAnalyzerPlugin.java
@@ -32,11 +32,6 @@ public class DummyAnalyzerPlugin extends Plugin {
         }
 
         @Override
-        public boolean recordProcessSuccessful() {
-            return true;
-        }
-
-        @Override
         public String name() {
             return "Dummy plugin";
         }
@@ -52,21 +47,6 @@ public class DummyAnalyzerPlugin extends Plugin {
 
         @Override
         public void stop() {
-        }
-
-        @Override
-        public void setPluginError(String exceptionType) {
-
-        }
-
-        @Override
-        public String getPluginError() {
-            return "";
-        }
-
-        @Override
-        public void freeResource() {
-
         }
     }
 }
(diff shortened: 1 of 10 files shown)


mir-am:
Thanks a lot for contribution and help. I will review the pull request this week. I need to test the FASTEN server with it and make necessary changes to the FASTEN server.

mir-am:
I have tested this branch. The FASTEN server and OPAL plug-ins work locally as expected. I will also test it on the Monster server.
Next, I will fix its conflicts with the master branch.

gousiosg:
So what is the verdict on this one? Are we going to merge or not?

mir-am:
> So what is the verdict on this one? Are we going to merge or not?

At the moment, we cannot merge this. I tested this branch. The solution is a bit simpler than what we have now in the master branch. However, I have some comments:
- As we agreed before, we have a topic for putting failed records for re-processing and debugging. However, It's not trivial with this solution to separate successful and failed records in the FASTEN server. Because we don't have the method `recordProcessSuccessful()` in this solution.

- With this strategy, we don't have the exact type of an Exception. For instance, for the case of FileNotFoundException, it only throws RunTimeException which is general and not helpful for re-processing records and debugging. Some Maven coordinates do not have JAR files and they cannot be re-processed.

- I think that methods like `setPluginError` and `getPluginError` help us a lot for re-processing records and debugging plug-ins. Since we will able to know what record and exception caused a plug-in to fail.

gousiosg:
The whole idea is that plug-ins should use exception wrapping (wrapping checked into unchecked exceptions to be handled by the plug-in runner) to avoid inconsistent handling of exceptions and extra interface methods.

ashkboos:
I have considered this policy on the new implementation. I think the only remained checked exception is during the URI generation, I will also remove that one. But it means we will fail to send the call graph even if we had problems with generating one method's URI.

gousiosg:
Closing due to inactivity

PR #6, opened 2020-02-14
Static analysis tools and .gitignore fix

ilyagrishkov:


mir-am:
@gousiosg, @ashkboos, @jhejderup 
I am okay to adhere to the Google coding conventions for FASTEN JAVA codebase and also having Maven Checkstyle plug-in.
What do the other contributors think about merging this pull request?

gousiosg:
Will this break CI if the committed code does not adhere to the style?

ilyagrishkov:
@gousiosg It shouldn't break CI now, unless we enforce it in CI config.

ilyagrishkov:
What is the conclusion on this PR? I think it would be helpful having it on master before merging both WALA and Metadata plugins, so we can be sure that both of them adhere to the same style conventions and also Mockito / Jacoco will help testing them.

mir-am:
> What is the conclusion on this PR? I think it would be helpful having it on master before merging both WALA and Metadata plugins, so we can be sure that both of them adhere to the same style conventions and also Mockito / Jacoco will help testing them.

I am okay with merging this PR.

@gousiosg, @ashkboos 
Are you two okay with merging this?

PR #8, opened 2020-02-21
Metadata Database Plugin

MihhailSokolov:
This plugin consumes Kafka topic and populates the metadata database.
(It is still work in progress)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@
 *.tar.gz
 *.rar
 
+*.iml
+
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 
@@ -169,7 +171,6 @@ fabric.properties
  .idea/modules.xml
  .idea/*.iml
  .idea/modules
- *.iml
  *.ipr
 
 ### Intellij Patch ###
@@ -180,3 +181,4 @@ fabric.properties
 
 # End of https://www.gitignore.io/api/intellij
 
+.idea/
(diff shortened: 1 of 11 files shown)


gousiosg:
Can you make the code more Java 11? For example, `Record record = new Record`  can be converted to `var record = new Record`

gousiosg:
Also, in `MetadataDao.java`, you should have methods that insert multiple records and return multiple IDs, for example:
```java
public List<long> insertCallables(long fileId, List<String> fastenUri, Timestamp createdAt, List<JSONObject> metadata)
```

In pure JDBC, you would use a prepared statement to add those; please use the equivalent in JOOQ.

MihhailSokolov:
> Can you make the code more Java 11? For example, `Record record = new Record` can be converted to `var record = new Record`

Done, thanks for the tip!

MihhailSokolov:
> Also, in `MetadataDao.java`, you should have methods that insert multiple records and return multiple IDs, for example:
> 
> ```java
> public List<long> insertCallables(long fileId, List<String> fastenUri, Timestamp createdAt, List<JSONObject> metadata)
> ```
> 
> In pure JDBC, you would use a prepared statement to add those; please use the equivalent in JOOQ.

Internally  jOOQ almost always uses prepared statements. However, at the moment jOOQ doesn't provide a nice method that would allow to add multiple records and get their IDs. There is a method called `batchInsert()` which can insert multiple records, but it doesn't return the IDs of those records. Since we really do need those IDs, I've decided for now just call multiple times already implemented method to add one records.
jOOQ has an open issue to implement this kind of functionality, so it will probably come out at some point in the future. Then we can change these methods.

mir-am:
Please continue the review and discussion here:
https://github.com/fasten-project/fasten/pull/11

PR #12, opened 2020-03-02
Extending revision call graphs

ashkboos:
This branch implemented:

1. Mapping nodes to CHA
2. Source file information per class
3. Metadata per edge for unresolved calls

Note: The new version of the revision call graph is [ExtendedRevisionCallGraph.java](https://github.com/fasten-project/fasten/blob/map_nodes_to_cha/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/callgraph/ExtendedRevisionCallGraph.java) which will be merged to the existing RevisionCallGraph.java if everyone confirms.


analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/callgraph/ExtendedRevisionCallGraph.java
```diff
@@ -34,214 +42,484 @@
...
+            MavenCoordinate.MavenResolver.downloadJar(coordinate.getCoordinate())
+                .orElseThrow(RuntimeException::new)
+        );
+
+        return new ExtendedRevisionCallGraph("mvn", coordinate.getProduct(),
+            coordinate.getVersionConstraint(), timestamp, partialCallGraph.getGENERATOR(),
+            MavenCoordinate.MavenResolver.resolveDependencies(coordinate.getCoordinate()),
+            partialCallGraph.getClassHierarchy(),
+            partialCallGraph.getGraph());
     }
```
ilyagrishkov: I like the new ExtendedRevisionCallGraph, but I have some concerns about this method. I don't think we should have a dependency from core to any analyser [in case this class is going to be moved to the core module]. This method depends on MavenCoordinate and PartialCallGraph from OPAL module. I think we should rather have a method in PartialCallGraph of each analyser that converts it to ExtendedRCG.

ashkboos: Thanks for the review, I moved it to the PartialCallGraph. Yes I agree, actually, in the beginning, it was there I moved it for cosmetic reasons. But anyway you are right it's better not to make core dependent on the analyzers.

PR #13, opened 2020-03-20
metadata DB connection and the deployment of the plug-in

mir-am:
In this PR, we work on the following tasks:

- [x] Add a method to the interface of KafkaConsumers for overriding the plug-ins' default topic.  
- [x] Create a `DBConnector` interface in the core module.
- [x] Move `codegen` classes from the `metadata-plugin` module to the `core` module.
- [x] Get DB credentials from the command line interface of the FASTEN server such as `DBUrl`, `username` and `password`.
- [x] Fix the deployment issue of the metadata DB plug-in, which means the plug-in should be detected by `PF4J` in the FASTEN server.
- [x] To change the default Kafka topic of plug-ins, read new topics from the CLI of the FASTEN server and override the default ones.


mir-am:
The plug-in developers (@MihhailSokolov and @ilyagrishkov), please add a test to your plug-in for `setTopic` method the same as the one I did for the OPAL plug-in https://github.com/fasten-project/fasten/pull/13/commits/ee0aba14fb0fd2a56d231d58f5aaec51b644fe85 or review your plug-in to make sure the `setTopic` method works as expected.

MihhailSokolov:
> The plug-in developers (@MihhailSokolov and @ilyagrishkov), please add a test to your plug-in for `setTopic` method the same as the one I did for the OPAL plug-in [ee0aba1](https://github.com/fasten-project/fasten/commit/ee0aba14fb0fd2a56d231d58f5aaec51b644fe85) or review your plug-in to make sure the `setTopic` method works as expected.

Done for metadata plugin

MihhailSokolov:
@mir-am, I moved `codegen` classes to `core`. Please check if you agree with where I put them. Feel free to move them somewhere else if you feel like they belong there better.

mir-am:
> @mir-am, I moved `codegen` classes to `core`. Please check if you agree with where I put them. Feel free to move them somewhere else if you feel like they belong there better.

For now, it is okay.

mir-am:
@MihhailSokolov 
You may want to use Mockito for this test https://github.com/fasten-project/fasten/pull/13/commits/c131bb85c6f2f4d6b0227c8db89a4d9865a0a1f3. Because I changed it to fix the build error.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java
```diff
@@ -58,30 +59,35 @@ public MetadataDatabasePlugin(PluginWrapper wrapper) {
...
+//         */
+//        public MetadataDBExtension(String callgraphTopic, DSLContext dslContext) {
+//            super();
+//            this.dslContext = dslContext;
+//            this.topic = callgraphTopic;
+//        }
+
+        @Override
+        public void getDBAccess(String DBUrl, String username, String password) throws SQLException {
+            PostgresConnector.getDSLContext(DBUrl, username, password);
```
MihhailSokolov: You should assign it to `dslContext` here. Otherwise, you don't really get a DSLContext required to interract with database. It should be like this:
```java
this.dslContext = PostgresConnector.getDSLContext(DBUrl, username, password);
```
And then it would make a bit more sense to call this method something like `setupDBAccess()` or simply `connectDB()`

mir-am: Okay. Thanks. I will fix it.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java
```diff
@@ -58,30 +59,35 @@ public MetadataDatabasePlugin(PluginWrapper wrapper) {
...
+//         * @param dslContext     DSLContext for jOOQ to query the database
+//         */
+//        public MetadataDBExtension(String callgraphTopic, DSLContext dslContext) {
+//            super();
+//            this.dslContext = dslContext;
+//            this.topic = callgraphTopic;
+//        }
+
+        @Override
+        public void getDBAccess(String DBUrl, String username, String password) throws SQLException {
```
MihhailSokolov: I think this method should also throw `IllegalArgumentException` since in `PostgresConnector.getDSLContext(...)` there is a check if DBUrl is in correct format and if it is not, it will throw `IllegalArgumentException`

MihhailSokolov:
> @MihhailSokolov
> You may want to use Mockito for this test [c131bb8](https://github.com/fasten-project/fasten/commit/c131bb85c6f2f4d6b0227c8db89a4d9865a0a1f3). Because I changed it to fix the build error.

I think there is no need for that. Most of the tests use mock of `MetadataDao` and after your changes we don't need to mock `DSLContext` and the tests pass in CI. So I think the part that you commented out in [c131bb8](https://github.com/fasten-project/fasten/commit/c131bb85c6f2f4d6b0227c8db89a4d9865a0a1f3) can be deleted.

MihhailSokolov:
@mir-am, Also should I remove a functionality to get database connection based `postgres.properties` file since we don't use that anymore?

mir-am:
> @mir-am, Also should I remove a functionality to get database connection based `postgres.properties` file since we don't use that anymore?

Yes. Please do it.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java
mir-am: I think it might be a better idea to catch `IllegalArgumentException` in the `getDBAccess` and set plug-in error rather than catching it at higher levels like the Main class.
Moreover, the project compiles successfully without throwing `IllegalArgumentException` in the `getDBAccess` and the IntelliJ also doesn't complain either!

mir-am:
@MihhailSokolov 
Okay. With recent changes, the plug-in now works as expected by running its JAR file outside the FASTEN server.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java
MihhailSokolov: It does not throw it because the URL we use is valid. But if user inputs invalid URL as a command-line argument it will throw uncaught `IllegalArgumentException`.
However, catching it in the `getDBAccess(..)` seems like a good solution too.

MihhailSokolov:
> @MihhailSokolov
> Okay. With recent changes, the plug-in now works as expected by running its JAR file outside the FASTEN server.

Nice! I will remove remove commented out code and add try/catch in the `getDBAccess(...)` to catch `IllegalArgumentException`

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java
```diff
@@ -118,6 +120,9 @@ public void consume(String topic, ConsumerRecord<String, String> record) {
                         }
                     });
                 } catch (Exception expected) {
+                    logger.error("An exception occurred {} ", expected.getStackTrace());
```
MihhailSokolov, line 121: There is no need for these 3 lines. An exception is already logged earlier in the transaction. It is just re-thrown to rollback the transaction, so we need an empty catch block to catch it. I will remove there 3 lines (121-123). Don't worry, this transactional error handling is quite confusing :)

mir-am: Okay, I will remove it.

analyzer/metadata-plugin/pom.xml
```diff
@@ -158,9 +164,9 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.1</version>
                 <configuration>
-                    <annotationProcessors>
-                        <annotationProcessor>org.pf4j.processor.ExtensionAnnotationProcessor</annotationProcessor>
```
mir-am, line 178: Please do not comment out the PF4J annotation processor in the POM file. If you have problem in Intellij regarding the annotation processor, please check out the last step in the main README file of the project.

MihhailSokolov: Got it. Thanks!

PR #17, opened 2020-04-06
Improved upsert strategy using constraints and batches

MihhailSokolov:
- [x] Implement new upsert strategy for `edges` using constraints on index
- [x] Append JSON metadata in `edges` instead of overwriting (if key exists, overwrite its value with new one, otherwise append new key-value pair)
- [x] Implement batch insertion for edges


gousiosg:
@mir-am Please merge

PR #20, opened 2020-04-26
Server refactoring

ilyagrishkov:
**The goal of this PR is implementing the following pipeline:**

1. A pkg release arrives in Kafka
2. The CG generator creates the CG and writes a DONE msg to its STDOUT topic
3. The metadata ingester plug-in writes to Postgres, updates the graph DB and writes a DONE msg to its STDOUT topic
4. All the downstream plugins are triggered by reading the DONE msg from the metadata ingester plug-in

**This PR's tasks are defined as follows:**
- [x] Design and implement the format of the DONE messages for the plug-ins.
- [x] Write DONE messages (events) for both kinds of plug-ins that have may or may not have a `KafkaProducer`. The DONE messages will be sent to plug-ins' STDOUT topics.
- [x] Write plug-ins errors to their STDERR topics.
- [x] Refactor server codebase.
- [x] Add support for more general purpose plugins that do not depend on Kafka consumers, e.g. security vulnerability ingester.


ilyagrishkov:
**A brief summary of changes made so far:**

1. Server creates out/err messages, receives optional payload from plugins, and sends those messages to a Kafka topic (with a standard name: _fasten.pluginname.out_ and _fasten.pluginname.err_);
2. Server support all types of Kafka plugins - consumer only, producer only, both consumer and producer;
3. A KafkaServerPlugin interface was added to allow for extensions of plugins types in the future, not just Kafka plugins;
4. KafkaConsumer & KafkaProducer were merged into KafkaPlugin and FastenKafkaConsumer & FastenKafkaProducer were merged into FastenKafkaPlugin;
5. Server codebase was refactored and documented;
6. Fasten Server can now be gracefully stopped if SIGINT is caught.

mir-am:
In general, the server refactoring is done well. The codebase is more neat and clean than the previous version. However, the new version is still experimental and needs extensive testing.

core/src/main/java/eu/fasten/core/plugins/KafkaPlugin.java
```diff
@@ -0,0 +1,16 @@
+package eu.fasten.core.plugins;
```
gousiosg, line 1: * No license header
* This class needs to be thoroughly documented

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/OPALPluginTest.java
```diff
@@ -35,125 +35,120 @@
 public class OPALPluginTest {
```
gousiosg, line 35: * Why are all the tests commented out?

core/src/main/java/eu/fasten/core/plugins/KafkaPlugin.java
```diff
@@ -0,0 +1,16 @@
...
+public interface KafkaPlugin<I, O> extends FastenPlugin {
+    Optional<List<String>> consumeTopics();
+
+    void setTopic(String topicName);
+
+    void consume(I record);
+
+    Optional<O> produce();
+
+    boolean recordProcessSuccessful();
```
gousiosg, line 15: What is the point of this method? If the plug-in creates an error (returned through the `getPluginError()` in the main class) then this will return `false`; else it will return `true`.

core/src/main/java/eu/fasten/core/plugins/FastenPlugin.java
```diff
@@ -34,57 +34,62 @@
...
-     * These two methods should be implemented so that the FASTEN server can retrieve the plug-in's error and its exception
-     * type. This will help to reprocess certain types of failed records.
+     * These two methods should be implemented so that the FASTEN server can retrieve
+     * the plug-in's error and its exception type. This will help to reprocess certain types of
+     * failed records.
+     *
+     * @param throwable exception or error encountered during plugin execution
      */
-    public void setPluginError(Throwable throwable);
+    void setPluginError(Throwable throwable);
```
gousiosg, line 79: I don't understand what the difference between `set` and `get` `PluginError` is

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -0,0 +1,363 @@
...
+     * Writes messages to server log and stdout/stderr topics.
+     *
+     * @param input input message [can be null]
+     * @param time  precessing time [can be null]
+     */
+    private void handleProducing(String input, String time) {
+        if (plugin.recordProcessSuccessful()) {
+            var result = plugin.produce();
+
+            emitMessage(this.serverLog, this.serverLogTopic,
```
gousiosg, line 148: Why are we writing server log messages to Kafka and not to the standard log? If the purpose is debugging, then we need to make a choice: we either write _everything_ to Kafka (e.g. by using a Kafka-aware logger) or we use a standard log aggregator that collects all server instance logs to a centralised location.

gousiosg: Moreover, this method cries for a `StringBuffer` to build the message string

gousiosg: The format should be JSON again, other wise we are inconsistent.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * This is a synchronous commits and will block until either the commit succeeds
+     * or an unrecoverable error is encountered.
+     */
+    private void doCommitSync() {
+        try {
+            connection.commitSync();
+            logger.debug("Committed the processed record...");
```
gousiosg, line 237: What is the point of this log message? What record? And why the ellipses at the end?

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -0,0 +1,363 @@
...
+        return "Unknown";
+    }
+
+    /**
+     * This is a utility method to get current committed offset of all partitions for a consumer.
+     *
+     * @param consumer Kafka consumer
+     * @param topics   list of topics
+     */
+    private void getOffsetForPartitions(KafkaConsumer<String, String> consumer,
```
gousiosg, line 291: This method looks unused in my IDE at least

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -0,0 +1,363 @@
...
+        // Note that this assumes that the consumer is subscribed to one topic only
+        for (PartitionInfo p : this.connection.partitionsFor(topics.get(0))) {
+            topicPartitions.add(new TopicPartition(topics.get(0), p.partition()));
+        }
+
+        ConsumerRecords<String, String> records = dummyPoll(this.connection);
+
+        if (records.count() != 0) {
+            for (TopicPartition tp : topicPartitions) {
+                logger.debug("Topic: {} | Current offset for partition {}: {}", topics.get(0),
```
gousiosg, line 328: Same as above: either log to Kafka or using the logger

gousiosg:
* As an additional comment, the cmd-line options in the server need to be split into groups with appropriate prefixes. For example, we can have `-du` for the database user and `-ku` for the Kafka user (if any). 

* The DB password must not be set at the command line (except if you know how to rewrite command lines), otherwise everybody will be able to read it by running `ps`. We need an environment variable (`$DBPASS`) for that.

* I am not sure what the `-f` option does and why we need it at the global server level.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
ilyagrishkov: I removed it

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
ilyagrishkov: Removed this log

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/OPALPluginTest.java
ilyagrishkov: I initially commented them all out to avoid fixing them while the new architecture was experimental, and completely forgot to fix them at the end. Now both OPAL and WALA plugins tests are fixed

core/src/main/java/eu/fasten/core/plugins/KafkaPlugin.java
ilyagrishkov: Added license headers to all classes and also documented this interface.

core/src/main/java/eu/fasten/core/plugins/KafkaPlugin.java
ilyagrishkov: This method was used only in place on the server. I removed it as it can be easily replaced in that one place.

core/src/main/java/eu/fasten/core/plugins/FastenPlugin.java
ilyagrishkov: I didn't get this comment. `set` method is used with each plugin to save exception that occurred in a variable. Then `get` method is used to retrieve this exception. Maybe it will be a good ides to make `get` method return Optional<Throwable>.

ilyagrishkov:
> * As an additional comment, the cmd-line options in the server need to be split into groups with appropriate prefixes. For example, we can have `-du` for the database user and `-ku` for the Kafka user (if any).
> * The DB password must not be set at the command line (except if you know how to rewrite command lines), otherwise everybody will be able to read it by running `ps`. We need an environment variable (`$DBPASS`) for that.
> * I am not sure what the `-f` option does and why we need it at the global server level.

I'm not sure what `-f` option are you referring to. Probably you've looked at MetadataPlugin parameters as it's the only place where I could find `-f` option. There is none on the server.

I have also removed `pass` parameter and now getting DB password from an environmental variable DBPASS. All other parameter got prefixes where necessary

ilyagrishkov:
> * As an additional comment, the cmd-line options in the server need to be split into groups with appropriate prefixes. For example, we can have `-du` for the database user and `-ku` for the Kafka user (if any).
> * The DB password must not be set at the command line (except if you know how to rewrite command lines), otherwise everybody will be able to read it by running `ps`. We need an environment variable (`$DBPASS`) for that.
> * I am not sure what the `-f` option does and why we need it at the global server level.

I'm not sure what `-f` option are you referring to. Probably you've looked at MetadataPlugin parameters as it's the only place where I could find `-f` option. There is none on the server.

I have also removed `pass` parameter and now getting DB password from an environmental variable DBPASS. All other parameter got prefixes where necessary

core/src/main/java/eu/fasten/core/plugins/FastenPlugin.java
gousiosg: Why do we need the `set` method? From an outsider perspective isn't `get` enough?

server/src/main/java/eu/fasten/server/FastenServer.java
```diff
@@ -54,7 +51,17 @@
             defaultValue = ".")
     private Path pluginPath;
 
-    @Option(names = {"-t", "--topic"},
+    @Option(names = {"-s", "--skip_offsets"},
```
gousiosg, line 54: Prefix `-s` with `k`, it is a Kafka related option

gousiosg, line 54: Since this is a Kafka-related option, prefix with `k` -> so it becomes `-ks`.

server/src/main/java/eu/fasten/server/FastenServer.java
```diff
@@ -54,7 +51,17 @@
...
+            defaultValue = "0")
+    private int skipOffsets;
+
+    @Option(names = {"-m", "--mode"},
+            description = "Deployment or Development mode")
+    private boolean deployMode;
+
+    @Option(names = {"-kt", "--topic"},
             paramLabel = "topic",
             description = "JSON file consists of topics for plug-ins.")
```
gousiosg, line 66: What does this mean?

gousiosg: Give default value and perhaps example

server/src/main/java/eu/fasten/server/connectors/PostgresConnector.java
```diff
@@ -16,34 +16,32 @@
...
      * @throws IllegalArgumentException if database URL has incorrect format and cannot be parsed
      */
-    public static DSLContext getDSLContext(String dbUrl, String user, String pass)
+    public static DSLContext getDSLContext(String dbUrl, String user)
             throws SQLException, IllegalArgumentException {
         if (!new Driver().acceptsURL(dbUrl)) {
             throw new IllegalArgumentException("Could not parse database URI: " + dbUrl);
         }
-        var connection = DriverManager.getConnection(dbUrl, user, pass);
+        var connection = DriverManager.getConnection(dbUrl, user, System.getenv("DBPASS"));
```
gousiosg, line 44: `DBPASS` is very generic. Perhaps rename to `FASTEN_DBPASS`. It could also read `PGPASSWORD`, the Postgres equivalent.

server/src/main/java/eu/fasten/server/FastenServer.java
gousiosg: Actually, why does this need to be a JSON string? JSON is very difficult to manipulate at the command line. Can we just provide a list of KV pairs, e.g. PLUGIN1:TOPIC_NAME1, PLUGIN2:TOPIC_NAME2 or even better allow `-kt` to repeat?

server/src/main/java/eu/fasten/server/FastenServer.java
```diff
@@ -54,7 +51,17 @@
             defaultValue = ".")
     private Path pluginPath;
```
gousiosg, line 52: Can we also add a `-pl`, `--list_plugins` option that reads displays a table with meta information about the loaded plug-ins? It can look like this:

```
PLUGIN, VERSION, DEFAULT_TOPIC, DESCR
OPAL, 0.0.1, fasten.mvn.packages, Reads a Maven coord and creates a callgraph
...
```

gousiosg:
> > * I am not sure what the `-f` option does and why we need it at the global server level.
> 
> I'm not sure what `-f` option are you referring to. Probably you've looked at MetadataPlugin parameters as it's the only place where I could find `-f` option. There is none on the server.

Yes, so

core/src/main/java/eu/fasten/core/plugins/FastenPlugin.java
ilyagrishkov: Technically we don't need it outside plugins, but I assume it was put there (which happened before I did any changes to the server) because every plugin uses this method to update its error field. I can delete it from the interface if needed.

server/src/main/java/eu/fasten/server/FastenServer.java
ilyagrishkov: This is very possible. I also though that there can be an argument to specify names of plugins that we want to run. So we don't have to maintain a list of different docker images with different plugins in each, we will just end up having one Docker image with all plugins, but when at the start of the container we'll specify plugins that need to run and the metadata for those will be automatically shown. So `-pl OPAL, MetadataDBExtension` will only run those two plugins and also automatically show metadata for them.

server/src/main/java/eu/fasten/server/connectors/PostgresConnector.java
ilyagrishkov: So it reads `FASTEN_DBPASS`, but if it cannot find it then `PGPASSWORD`. Something like that or only one of them?

server/src/main/java/eu/fasten/server/FastenServer.java
gousiosg: Yes, makes sense. We will have a single Docker image in any case after @wzorgdrager's changes. Let's do that.

server/src/main/java/eu/fasten/server/connectors/PostgresConnector.java
gousiosg: Exactly that!

core/src/main/java/eu/fasten/core/plugins/FastenPlugin.java
gousiosg: Sure, but in interfaces we only need to have methods that outside clients need and just them. Clients should not care about how plug-ins do set their error messages. I 'd remove `setPluginError`

ilyagrishkov: Removed

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
ilyagrishkov: The only things that are written to Kafka are stdout and stderr now. Using logger for the rest

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
ilyagrishkov: Switched everything to logger

server/src/main/java/eu/fasten/server/FastenServer.java
ilyagrishkov: It's `-ks` now

server/src/main/java/eu/fasten/server/FastenServer.java
ilyagrishkov: I implemented key-value pairs, but instead of `:` I'm using `=` which is a way `picocli` is handling maps, which seems like a natural way of handling things such as this one. So we'll have OPAL=fasten.maven.pkg for example.

server/src/main/java/eu/fasten/server/connectors/PostgresConnector.java
ilyagrishkov: Implemented

server/src/main/java/eu/fasten/server/FastenServer.java
ilyagrishkov: Now we can specify a number of plugins to run using `-pl` (can be repeated). Adding `-pl OPLA` will end up stopping and unloading all other plugins and printing OPAL's metadata to log. The only difference is that default_topic is not included in this metadata, as it's a generic way of printing info for FastenPlugins, but topic is KafkaPlugins-specific

PR #22, opened 2020-04-29
Graph Plugin

MihhailSokolov:
This Pull Request implements new architecture for updating graph database:
- [x] Change Metadata Plugin to produce GID Graph to the Kafka topic
- [x] Create new plugin called Graph Plugin that will consume topic with GID graphs and populate graph database (RocksDB)

Below is the example of the GID Graph that is used to communicate between Metadata and Graph Plugins:
```json
{
    "index": 1,
    "product": "test",
    "version": "0.0.1",
    "nodes": [0, 1, 2],
    "numInternalNodes": 2,
    "edges": [[0, 1], [1, 2]]
}
```


core/src/main/java/eu/fasten/core/plugins/GraphDBConnector.java
```diff
@@ -0,0 +1,34 @@
...
+ * A plug-in that needs to access graph database (RocksDB)
+ */
+public interface GraphDBConnector extends FastenPlugin {
+
+    /**
+     * Sets location where RocksDB database is stored
+     *
+     * @param dir Path to database directory
+     */
+    void setRocksDbDir(String dir) throws RocksDBException;
```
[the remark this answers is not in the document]
gousiosg: Moreover, I remember we discussed that we don't do checked exceptions.

core/src/main/java/eu/fasten/core/plugins/GraphDBConnector.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package eu.fasten.core.plugins;
+
+import org.rocksdb.RocksDBException;
+
+/**
+ * A plug-in that needs to access graph database (RocksDB)
+ */
+public interface GraphDBConnector extends FastenPlugin {
```
gousiosg, line 26: This interface does not look very useful. A graph connector should return a reference to an initialised `GraphDBDAO` to plug-ins so that the plug-in can directly query, same as the PostgresConnector.  A plug-in should not care about the location of the underlying database.

core/src/main/java/eu/fasten/core/plugins/GraphDBConnector.java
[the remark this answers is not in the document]
MihhailSokolov: I was a bit confused about error handling. When initializing RocksDao it can throw `RocksDBException`. Should the try/catch block be in the Graph Plugin class or in the server when initializing the plugin?

core/src/main/java/eu/fasten/core/plugins/GraphDBConnector.java
MihhailSokolov: I can change it to `setRocksDao(RocksDao rocksDao)` and then we can either create RocksDao instance in Fasten Server when initializing plugin or have a separate class RocksDBConnector with one static method that will take directory and return RocksDao instance (quite similar to PostgresConnector). Which one should I implement?

gousiosg: The later. Is RocksDB protecting itself (e.g. at the filesystem level) against multiple writers? If not, we need to design for that.

core/src/main/java/eu/fasten/core/plugins/GraphDBConnector.java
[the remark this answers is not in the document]
gousiosg: You should wrap RocksDBException into a Runtime exception and let if flow to the server.

core/src/main/java/eu/fasten/core/plugins/GraphDBConnector.java
MihhailSokolov: Hm, now that I think of it I am not sure that `setRocksDao(...)` would be appropriate. Since RocksDao is located in graph plugin module, we would have to have a dependency to graph plugin from core or server. I am not sure If that's a good solution.

MihhailSokolov: In order to solve the issue with circular dependency, I could move `RocksDao` to core. Do you think that would be appropriate @gousiosg ?

gousiosg: Yes

core/src/main/java/eu/fasten/core/plugins/GraphDBConnector.java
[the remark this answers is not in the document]
MihhailSokolov: Done

gousiosg:
LGTM

vigna:
LGTM

PR #26, opened 2020-05-25
Saving call graphs

ilyagrishkov:
Adds an ability to specify a path when starting a FASTEN server where call graphs generated by plugins implementing a marker interface CallGraphGeneratorPlugin will be stored in the JSON format.

TODO:

- [x] Add ability to specify a path where call graphs will be saved.
- [x] Identify which plugins' output needs to be stored.
- [x] Come up with a directory structure for saving call graphs.


server/src/main/java/eu/fasten/server/FastenServer.java
```diff
@@ -60,6 +60,16 @@
...
     List<String> plugins;
 
+    @Option(names = {"-po", "--plugin_output"},
+            paramLabel = "dir",
+            description = "Path to directory with call graphs")
+    Map<String, String> outputDirs;
+
+    @Option(names = {"-pol", "--plugin_output_link"},
+            paramLabel = "dir",
+            description = "Path to directory with call graphs")
```
gousiosg, line 70: We are not only saving callgraphs

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -142,16 +152,29 @@ private void handleConsuming() {
...
-        if (plugin.getPluginError() == null) {
+        try {
+            if (plugin.getPluginError() != null) {
+                throw plugin.getPluginError();
+            }
+
             var result = plugin.produce();
+            String payload = null;
+            if (result.isPresent()) {
+                if (writeDirectory != null && !writeDirectory.equals("")) {
```
gousiosg, line 163: Better call `result.get()` before this `if` and store it in a variable. More debuggable and you can loose the `else`

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -176,6 +199,35 @@ private void emitMessage(KafkaProducer<String, String> producer, String topic, S
...
+     * Writes {@link RevisionCallGraph} to JSON file and return JSON object containing
+     * a link to to written file.
+     *
+     * @param result String of JSON representation of {@link RevisionCallGraph}
+     * @return Path to a newly written JSON file
+     */
+    private String writeToFile(String result)
+            throws IOException, NullPointerException {
+        var path = plugin.getOutputPath();
+        var pathWithoutFilename = path.substring(0, path.lastIndexOf("/"));
```
gousiosg, line 212: You are assuming a unix filesystem here; this will probably not work on windows. Please make cross-platform

server/src/main/java/eu/fasten/server/FastenServer.java
ilyagrishkov: Forgot to change it. Fixed now

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
ilyagrishkov: Made this part more debuggable and overall cleaned it a bit

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
ilyagrishkov: Replaced all "/" that I could find with File.separator, so now it should work on every platform.

ilyagrishkov:
Here is a list of change in this pull request:

- Support for saving plugins output messages to a specified directory. New CLI commands on the server are `-po` and `-pol` that are used to specify plugin output path and plugin link respectively. Both command have the following format: `plugin_name=path_or_link`
- Properly catching shutdown hook and made Kafka consumers stop gracefully, so there is no waiting time for rejoining a group on a server restart. The server doesn't terminate immediately, but it finishes producing a few coordinates it already has consumed.
- Each plugin overrides an `outputPath` method that constructs a plugin-specific directory hierarchy where output message will be written.
- Changed a separator between groupId and artifactId in RevisionCallGraph both in product and dependencies, so those parts can be easily extracted. 
- Changed MavenCoordinate repo from a hardcoded link to a list of link that can be updated when running Main classes of call graph generator plugin. Closes #30 
- Removed JRE dependency for WALA, so it runs on every platform. Closes #32

analyzer/base-analyzer/src/main/java/eu/fasten/analyzer/baseanalyzer/MavenCoordinate.java
```diff
@@ -276,7 +276,7 @@ public String toPomUrl(String repo) {
 
                 final var dependency = new RevisionCallGraph.Dependency(
                         "mvn",
-                        groupId + "." + artifactId,
+                        groupId + ":" + artifactId,
```
gousiosg, line 279: I don't think this works on Windows. See this: https://en.wikipedia.org/wiki/Filename#In_Windows

ilyagrishkov: It won't be used as a filename. It's just for the purpose of being able to differentiate groupId and artifactId in product name and dependency name.

gousiosg: Ok, let's merge then

PR #27, opened 2020-05-28
Repo cloning plugin

MihhailSokolov:
Creates a new plugin called RepoCloner which consumes a topic with maven coordinates with repository URL, clones the repository (if repository already exists, just performs `git pull`) into a hierarchical directory structure and produces maven coordinates with the path to the repository on the disk.
Below is the example of input and output of the plugin:
* Input
```json
{
	"payload": {
		"artifactId": "junit",
		"groupId": "junit",
		"version": "4.11",
		"repoUrl": "https://github.com/junit-team/junit4.git"
	}
}
```
_P.S. All attributes like `artifactId`, `groupId`, etc are optional._

* Output
```json
{
	"artifactId": "junit",
	"groupId": "junit",
	"version": "4.11",
	"repoPath": "<base dir>/j/junit-team/junit4"
}
```
In general hierarchical structure looks like this:
`<base dir>/<first letter of repo owner>/<repo owner>/<repo name>`

In order to use the plugin with FASTEN server, `-b <baseDir>` argument must be provided to the server.


gousiosg:
Why is this plug-in so focused on Maven projects?

MihhailSokolov:
> Remove commented out pieces of code

I will remove some commented out code from `pom.xml` once a small problem with dependencies is fixed

MihhailSokolov:
> Why is this plug-in so focused on Maven projects?

I am still not sure how the hierarchy should look like. I can change it to take in the forge from the consumed record, the same way as it takes in artifactId, groupId, version, etc.

gousiosg:
We will need to clone repos for multiple languages; right now we assume that the input information contains Maven related information. How can we replicate this information in the case of, e.g., Python

MihhailSokolov:
> We will need to clone repos for multiple languages; right now we assume that the input information contains Maven related information. How can we replicate this information in the case of, e.g., Python?

I will add functionality to retrieve `forge` from the consumed record and I can replace `artifactId` and `groupId` (which is quite specific to Java) with `product`. In the case of Java, `product` will be `<groupId>.<artifactId>`. And then for other languages, it can be simply the name of the product. However, I am not sure how the hierarchy will be built then because currently, it is quite dependent on knowing what is the artifact and what is the group.
Another idea is that we can extract that kind of information from GitHub URLs, e.g. `https://github.com/fasten-project/fasten` -> `group='fasten-project', artifact='fasten'` and can be stored at `<base_dir>/<first letter of group>/<group>/<artifact>` making it quite similar to GitHub's structure.

MihhailSokolov:
@gousiosg So how should I implement the hierarchy for repositories and JAR files? Above are the two options I see for repositories. And for JAR files something similar could be done.

analyzer/javacg-wala/pom.xml
```diff
@@ -67,22 +67,22 @@
...
-            <artifactId>slf4j-api</artifactId>
-            <version>1.7.30</version>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-engine</artifactId>
-            <version>5.4.0</version>
-            <scope>test</scope>
-        </dependency>
+<!--        <dependency>-->
```
gousiosg, line 70: Please remove dead code

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/Main.java
```diff
@@ -0,0 +1,65 @@
...
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "RepoCloner")
+public class Main implements Runnable {
+
+    private static final Logger logger = LoggerFactory.getLogger(Main.class);
+
+    @CommandLine.Option(names = {"-f", "--file"},
+            paramLabel = "JSON File",
+            description = "Path to JSON file which contains maven coordinate, repoUrl and jarUrl")
```
gousiosg, line 36: Why maven coordinate? How are we going to process Python projects?

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/Main.java
```diff
@@ -0,0 +1,65 @@
...
+    String jsonFile;
+
+    @CommandLine.Option(names = {"-d", "--base-dir"},
+            paramLabel = "Directory",
+            description = "Path to base directory where repo hierarchy must be crated")
```
gousiosg, line 41: s/must be/will be/

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/Main.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public void run() {
+        var repoCloner = new RepoClonerPlugin.RepoCloner();
+        repoCloner.setBaseDir(baseDir);
+        final FileReader reader;
+        try {
+            reader = new FileReader(jsonFile);
+        } catch (FileNotFoundException e) {
+            logger.error("Could not find the JSON file at " + jsonFile, e);
```
gousiosg, line 57: Change message to: "Could not find input JSON " + jsonFile

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
```diff
@@ -0,0 +1,163 @@
...
+
+        @Override
+        public String name() {
+            return "Repo Cloner Plugin";
+        }
+
+        @Override
+        public String description() {
+            return "Repo Cloner Plugin. "
+                    + "Consumes GitHub repository URL (if present), "
```
gousiosg, line 136: It does not consume a GitHub repo URL, does it?

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/Main.java
MihhailSokolov: Not sure I understand what you mean🙂

MihhailSokolov: I see now😁
I will change it

MihhailSokolov:
All commented out code has been deleted as far as I know (maybe I'm wrong), I will check again against your comments. I will change everything I can.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/Main.java
gousiosg: I hope you know how to use `sed` :-)

analyzer/javacg-wala/pom.xml
MihhailSokolov: This is not dead code. It is commented out because it interferes with the server dependencies but if you want to run it as a standalone plugin, these dependencies are needed. So we just have to uncomment them whenever we want to run Main class of the plugin and they are commented out the rest of time.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
MihhailSokolov: It can consume any repository URL which ends with something like `/<repo_owner>/<repo_name>`

MihhailSokolov:
@gousiosg I have made all attributes in input record JSON optional.
We need to have `artifactId`, `groupId` optional because projects with other languages might prefer different naming. Plus, they are not really needed as all relevant information on where to actually clone the repo is extracted from `repoUrl`.
`version` is optional because it is also not really needed, as repository files are not related to a certain version.
`repoUrl` is optional because for some project we won't have a repository URL.
I have addressed all your other comments too. Sorry for the delay :)

PR #42, opened 2020-07-28
POM cleaning 

ilyagrishkov:
Cleaning up duplicate dependencies, creating proper dependency tree, and structuring pom.xml-s


mir-am:
@ilyagrishkov 
Thanks for the maintenance work.
We need to make sure that the following concerns are okay:
- All the FASTEN plug-ins can successfully be loaded and process records as usual.
- The size of plug-ins' JAR file are not increased because of adding core dependencies. However, I've seen that you excluded `slf4j` in `<exclude>` tags to solve the JAR size issue.

ilyagrishkov:
> Why are the RepoCloner tests disabled?

I've discussed those tests with @MihhailSokolov as they were taking too long to be executed, and I turned them off until he changes some of them to use Mockito to reduce execution time. Some RepoCloner tests will remain disabled, but they can be manually run on a local machine if needed.

ilyagrishkov:
> @ilyagrishkov
> Thanks for the maintenance work.
> We need to make sure that the following concerns are okay:
> 
> * All the FASTEN plug-ins can successfully be loaded and process records as usual.
> * The size of plug-ins' JAR file are not increased because of adding core dependencies. However, I've seen that you excluded `slf4j` in `<exclude>` tags to solve the JAR size issue.

So far plugins can be loaded and started by the server without any problems, but the I've only fully run one of them. There will be more testing before merging this PR. 
With regard to JAR sizes - some of them increased slightly (about 1MB bigger), some remained the same size, and  repoCloner plugin decreased in size by 3 MB. Overall the size is approximately the same.

analyzer/metadata-plugin/pom.xml
```diff
@@ -77,20 +42,25 @@
                         </goals>
                         <configuration>
                             <minimizeJar>false</minimizeJar>
-                            <keepDependenciesWithProvidedScope>false</keepDependenciesWithProvidedScope>
+                            <keepDependenciesWithProvidedScope>false
+                            </keepDependenciesWithProvidedScope>
```
proksch, line 46: This is another case, where line breaks look super weird after the PR.

core/pom.xml
```diff
@@ -73,49 +87,55 @@
...
+            <!--                            <includes>.*</includes>-->
+            <!--                            <inputSchema>public</inputSchema>-->
+            <!--                        </database>-->
+            <!--                        <target>-->
+            <!--                            <packageName>eu.fasten.core.data.metadatadb.codegen</packageName>-->
+            <!--                            <directory>src/main/java</directory>-->
+            <!--                        </target>-->
+            <!--                    </generator>-->
+            <!--                </configuration>-->
+            <!--            </plugin>-->
```
proksch, line 138: Do we still need this part of the config? Instead of commenting it out, future us could also just look it up in the repository history.

core/src/test/java/eu/fasten/core/data/RevisionCallGraphTest.java
```diff
@@ -1,209 +1,204 @@
...
+        assertEquals(3, constraints.size());
+        assertEquals("3.1", constraints.get(0).lowerBound);
+        assertEquals("7.1", constraints.get(0).upperBound);
+        assertEquals("9", constraints.get(1).lowerBound);
+        assertEquals("9", constraints.get(1).upperBound);
+        assertEquals("10.3", constraints.get(2).lowerBound);
+        assertNull(constraints.get(2).upperBound);
+
+        assertEquals(new JSONArray(spec.replaceAll(" ", "")).toString(), RevisionCallGraph.Constraint.toJSON(constraints).toString());
+    }
```
proksch, line 60: Would it make sense to split this test into multiple smaller tests? All individual blocks seem to be unrelated.

core/src/test/java/eu/fasten/core/data/RevisionCallGraphTest.java
```diff
@@ -1,209 +1,204 @@
...
+                "]";
+        JSONArray json = new JSONArray(spec);
+        List<List<RevisionCallGraph.Dependency>> depset = RevisionCallGraph.Dependency.depset(json);
+        RevisionCallGraph.Dependency d = depset.get(0).get(0);
+        assertEquals("maven", d.forge);
+        assertEquals("foo.bar", d.product);
+        List<RevisionCallGraph.Constraint> constraints = d.constraints;
+        assertEquals(3, constraints.size());
+        assertEquals("3.1", constraints.get(0).lowerBound);
+        assertEquals("7.1", constraints.get(0).upperBound);
```
proksch, line 103: Does `Constraint` have hashCode and equals implemented? If yes, this type of test could be reduced to a single assert between `new Constraint(...)` and `constraints.get(0)`, avoiding two asserts for the two fields (the same is true for the following fields as well). There might be a chance to reduce the size of this test suite.

core/src/test/java/eu/fasten/core/data/RevisionCallGraphTest.java
```diff
@@ -1,209 +1,204 @@
...
+                "         \"//b/their.package/TheirClass.method(TheirOtherClass)TheirOtherClass\",\n" +
+                "         {\n" +
+                "           \"invokevirtual\":\"1\"\n" +
+                "         }\n" +
+                "       ]\n" +
+                "     ]\n" +
+                "   }\n" +
+                "}";
+        JSONObject json = new JSONObject(callGraph);
+        RevisionCallGraph cg = new RevisionCallGraph(json);
```
proksch, line 176: This is an interesting construct... why is the `JSONObject` required to instantiate the `RevisionCallGraph`? Wouldn't it be better to leave the parsing logic outside of the data structure and better use, for example, `GSON` or `Jackson` for the deserialization of the JSON string?

core/src/test/java/eu/fasten/core/data/metadatadb/MetadataDaoTest.java
```diff
@@ -342,21 +374,23 @@ public void insertDependencyTest() {
         long packageId = 8;
         long dependencyId = 42;
         var versionRange = new String[]{"1.0.0-1.9.9"};
-        var insertValues = Mockito.mock(InsertValuesStep3.class);
+        var insertValues = Mockito.mock(InsertValuesStep4.class);
```
proksch, line 377: Is this a good class name? I am not sure what "step 4" refers to, but it seems that this is going to change again, every time there will be another step.

core/src/test/java/eu/fasten/core/data/metadatadb/MetadataDaoTest.java
```diff
@@ -366,24 +400,26 @@ public void insertMultipleDependenciesTest() throws IllegalArgumentException {
...
-        var insertValues = Mockito.mock(InsertValuesStep3.class);
+        var metadata = Arrays.asList(new JSONObject(), new JSONObject());
+        var insertValues = Mockito.mock(InsertValuesStep4.class);
         Mockito.when(context.insertInto(Dependencies.DEPENDENCIES, Dependencies.DEPENDENCIES.PACKAGE_VERSION_ID,
-                Dependencies.DEPENDENCIES.DEPENDENCY_ID, Dependencies.DEPENDENCIES.VERSION_RANGE)).thenReturn(insertValues);
-        Mockito.when(insertValues.values(packageId, dependencyIds.get(0), versionRanges.get(0))).thenReturn(insertValues);
-        Mockito.when(insertValues.values(packageId, dependencyIds.get(1), versionRanges.get(1))).thenReturn(insertValues);
+                Dependencies.DEPENDENCIES.DEPENDENCY_ID, Dependencies.DEPENDENCIES.VERSION_RANGE, Dependencies.DEPENDENCIES.METADATA)).thenReturn(insertValues);
+        Mockito.when(insertValues.values(packageId, dependencyIds.get(0), versionRanges.get(0), JSONB.valueOf("{}"))).thenReturn(insertValues);
+        Mockito.when(insertValues.values(packageId, dependencyIds.get(1), versionRanges.get(1), JSONB.valueOf("{}"))).thenReturn(insertValues);
```
proksch, line 408: The additional parameters are unrelated to the clean-up, are they? What changed here?

analyzer/graph-plugin/pom.xml
```diff
@@ -70,43 +42,49 @@
...
-<!--                                    <exclude>net.sf.jung:jung-io:*</exclude>-->
-<!--                                    <exclude>net.sf.jung:jung-algorithms:*</exclude>-->
-<!--                                    <exclude>com.jcraft:jsch:*</exclude>-->
-<!--                                    <exclude>com.jcraft:jzlib:*</exclude>-->
+                                    <!--<exclude>com.martiansoftware:jsap:*</exclude>-->
+                                    <!--<exclude>it.unimi.dsi:dsiutils:*</exclude>-->
+                                    <!--<exclude>it.unimi.dsi:fastutil:*</exclude>-->
+                                    <!--<exclude>org.codehaus.mojo:animal-sniffer-annotations:*</exclude>-->
+                                    <!--<exclude>it.unimi.dsi:webgraph:*</exclude>-->
+                                    <!--<exclude>it.unimi.dsi:sux4j:*</exclude>-->
```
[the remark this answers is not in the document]
ilyagrishkov: They were indeed unnecessary. I removed them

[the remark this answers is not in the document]
ilyagrishkov: They were used when fixing plugin compatibility issues with the server. Now when everything works there is no need for them.

core/src/test/java/eu/fasten/core/data/metadatadb/MetadataDaoTest.java
ilyagrishkov: @MihhailSokolov

analyzer/graph-plugin/pom.xml
```diff
@@ -70,43 +42,49 @@
...
+                                    <!--<exclude>net.sf.jung:jung-io:*</exclude>-->
+                                    <!--<exclude>net.sf.jung:jung-algorithms:*</exclude>-->
+                                    <!--<exclude>com.jcraft:jsch:*</exclude>-->
+                                    <!--<exclude>com.jcraft:jzlib:*</exclude>-->
                                 </excludes>
                             </artifactSet>
                             <transformers>
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                                <transformer
+                                        implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
```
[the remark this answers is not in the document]
ilyagrishkov: It's a line length limit in my IDE, not project specific. All of those broken down lines were caused by auto-formatting. Now I increased allowed length and fixed all such cases, so it shouldn't be a problem in the future.

analyzer/metadata-plugin/pom.xml
ilyagrishkov: Fixed

core/src/test/java/eu/fasten/core/data/metadatadb/MetadataDaoTest.java
ilyagrishkov: @MihhailSokolov

core/src/test/java/eu/fasten/core/data/RevisionCallGraphTest.java
ilyagrishkov: Now when I think about it, refactoring the RCG and it's tests seems like a nice idea for the next PR, especially because (as far as I'm know) our partners are planning to create language-specific RCGs, so we need a good base class.

core/src/test/java/eu/fasten/core/data/metadatadb/MetadataDaoTest.java
MihhailSokolov: I didn't choose that name unfortunately :)
It's a class from jOOQ and `4` refers to the number of attributes that will be inserted in the database.
The number of attributes increased from 3 to 4 because we added dependency metadata.
So we can't really do anything about this naming.

core/src/test/java/eu/fasten/core/data/metadatadb/MetadataDaoTest.java
MihhailSokolov: There was an error in the test. This test was built to test the functionality of the insertion of 3 attributes but recently we added the fourth attribute but forgot to update the test. So that's what was done here.

core/src/test/java/eu/fasten/core/data/RevisionCallGraphTest.java
ilyagrishkov: Will be taken into account in the RCG refactoring PR

core/src/test/java/eu/fasten/core/data/RevisionCallGraphTest.java
ilyagrishkov: Will be added when refactoring RCG

core/pom.xml
ilyagrishkov: We are using it to generate JOOQ classes whenever the DB schema changes. So it better be left commented out for the sake of convenience.

proksch: Ok, makes sense

ilyagrishkov:
@mir-am With the support of @MihhailSokolov all the plugins have been tested on the server and all problems were fixed.

PR #44, opened 2020-08-03
Improve RepoCloner

MihhailSokolov:
New features:
- [x] Clone with SSH links from GitHub
- [x] Clone with HTTP links from BitBucket (both Mercurial and Git repositories)
- [x] Clone with SSH links from BitBucket (both Mercurial and Git repositories)
- [x] Clone with HTTP links from Apache (Subversion repositories)
- [x] Investigate other repository hosts 

The functionality has been tested on 500 URLs from the Metadata Database and success rate was reported to be ~77%.

### DIFF
--- a/analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
+++ b/analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
@@ -23,16 +23,16 @@ import eu.fasten.analyzer.repoclonerplugin.utils.HgCloner;
 import eu.fasten.analyzer.repoclonerplugin.utils.SvnCloner;
 import eu.fasten.core.plugins.DataWriter;
 import eu.fasten.core.plugins.KafkaPlugin;
+import java.io.File;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
 import org.json.JSONObject;
 import org.pf4j.Extension;
 import org.pf4j.Plugin;
 import org.pf4j.PluginWrapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.io.File;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
 
 public class RepoClonerPlugin extends Plugin {
 
@@ -122,7 +122,8 @@ public class RepoClonerPlugin extends Plugin {
                     + (artifact == null ? "" : artifact.charAt(0) + File.separator)
                     + artifact + File.separator
                     + (product == null ? "unknown" : product.replace(":", "_")) + ".json";
-            var repoUrl = json.optString("repoUrl").replaceAll("[\\n\\t ]", "");;
+            var repoUrl = json.optString("repoUrl").replaceAll("[\\n\\t ]", "");
+            ;
             if (!repoUrl.isEmpty()) {
                 var gitCloner = new GitCloner(baseDir);
                 var hgCloner = new HgCloner(baseDir);
@@ -144,7 +145,7 @@ public class RepoClonerPlugin extends Plugin {
         }
 
         public boolean cloneRepo(String repoUrl, String artifact, String group,
-                              GitCloner gitCloner, HgCloner hgCloner, SvnCloner svnCloner) {
+                                 GitCloner gitCloner, HgCloner hgCloner, SvnCloner svnCloner) {
             var cloned = false;
             if (repoUrl.startsWith("scm:git:") || repoUrl.startsWith("scm:svn:")) {
                 repoUrl = repoUrl.substring(8);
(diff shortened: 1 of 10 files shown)


ilyagrishkov:
You only changed version in the RepoClonerPlugin, but it's still 0.0.1-SNAPSHOT in the pom.xml. I think those versions should match

MihhailSokolov:
> You only changed version in the RepoClonerPlugin, but it's still 0.0.1-SNAPSHOT in the pom.xml. I think those versions should match

Thanks, I will change that.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
```diff
@@ -19,20 +19,20 @@
...
 import org.json.JSONObject;
 import org.pf4j.Extension;
 import org.pf4j.Plugin;
 import org.pf4j.PluginWrapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.io.File;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
```
ilyagrishkov, line 35: Those imports don't follow the lexicographical order.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
```diff
@@ -116,37 +116,94 @@ public void consume(String record) {
...
                 }
                 if (getPluginError() == null) {
                     logger.info("Cloned repository"
-                            + (product == null ? "" : "of '" + product + "'")
+                            + (product == null ? "" : " of '" + product + "'")
                             + " from " + repoUrl + " to " + repoPath);
                 }
             } else {
                 logger.info("Repository URL not found");
             }
```
ilyagrishkov, line 143: I don't think that RuntimeException is descriptive enough in this case. You catch all exceptions within `cloneRepo([...])` method and set `cloned` to either true or false. I would suggest throwing those exceptions from the method and catch Throwable in `consume` method. This way we will have different exceptions per different repo types and they will be more descriptive, compared to only having RuntimeExceptions.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
```diff
@@ -116,37 +116,94 @@ public void consume(String record) {
...
         }
 
-        public void cloneRepo(String repoUrl, GitCloner gitCloner)
-                throws GitAPIException, IOException {
-            repoPath = gitCloner.cloneRepo(repoUrl);
+        public boolean cloneRepo(String repoUrl, String artifact, String group,
```
ilyagrishkov, line 146: Linked with my previous comment: you could make this method `void` and get rid of `cloned` variable if you delegate catching exceptions to the `consume` method.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
MihhailSokolov: It's a bit more complicated than that. The thing is that it is sometimes hard to tell which kind of repository is it by the URL. For example, BitBucket hosts both Git and Mercurial repositories, and their URLs would be exactly the same. So if one of the cloners (GitCloner for example) gives an error, it doesn't mean that we can't clone the repository at all. It means that it might also be simple not Git repository and we need to try all other cloners.
I am planning to do some refactoring and improve error handling in the next version of RepoCloner.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
MihhailSokolov: Since refactoring of error handling will be done in the next version, there are no exceptions to catch, so no need to change this at the moment.

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
MihhailSokolov: Will fix that

analyzer/repo-cloner-plugin/src/main/java/eu/fasten/analyzer/repoclonerplugin/RepoClonerPlugin.java
ilyagrishkov: Fair enough

Issue #47, opened 2020-08-06
Some metadataDB tests mock their target functions in their entirety


marcomicera:
As an example, the following test
https://github.com/fasten-project/fasten/blob/b674bd3155950fba9c2a7bed2816cce7855e3974/core/src/test/java/eu/fasten/core/data/metadatadb/MetadataDaoTest.java#L89-L117

mocks the entire behavior of its target function:
https://github.com/fasten-project/fasten/blob/b674bd3155950fba9c2a7bed2816cce7855e3974/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java#L59-L82

After a brief yet super-productive call with @MihhailSokolov we figured out that these tests originally targetted the data access layer, hence mocking the database layer. Due to the lack of the first, a lot of tests ended up targeting an entirely mocked-up version of their corresponding target function.

What I would personally do is to mimic an actual insertion of data in a temporary (test) database and consequently write some assertions relative to it, but this sounds very much like an integration test that might already be present. Either way, I would get rid of these tests.

MihhailSokolov:
I agree that these tests are quite useless as they mock the complete behavior of target functions. The only their use is to basically double-check that queries are correct, however, it makes it quite hard to maintain them as you'd need to rewrite a lot of big tests whenever you change a query. I guess I will delete them if nobody minds.
I also looked into the [article](https://blog.jooq.org/2013/02/20/easy-mocking-of-your-database/) that @fullsushidev sent during the call. It won't help much in the case of previously mentioned tests as they are simply insertions in the database. But I do think it might be useful for you @marcomicera in the RestAPI plugin for mocking the database when you test retrieving data.

PR #49, opened 2020-08-06
Delete MetadataDaoTest class

MihhailSokolov:
## Description
Deletes `MetadataDaoTest` class

## Motivation and context
Tests in `MetadataDaoTest` class are basically useless as they mock the whole functionality of tested functions.
Targeted test class `MetadataDao` consists of several methods which are just jOOQ queries (no logic). Since those queries are mocked, these tests actually don't test anything. They simply replicate the same behavior but mocked.
Fixes issue #47 

## Testing
No testing was needed since it was simply a deletion of the test class.

## Task list  
- [x] Delete `MetadataDaoTest` class

## Additional context
Problem was noticed and discussed with @marcomicera and @fullsushidev


mir-am:
@MihhailSokolov 
What is final verdict for this PR?
Can we merge this?

MihhailSokolov:
> @MihhailSokolov 
> What is final verdict for this PR?
> Can we merge this?

I would say yes.
I still think that these tests are kind of pointless as Endocode pointed out. So if nobody minds, we should delete them.
We can still discuss it more on the meeting if needed.

PR #51, opened 2020-08-12
Improve POM Analyzer

MihhailSokolov:
## Description
This PR improves the functionality of the POM Analyzer, fixes a bug with duplicate package names in the database, and adds extractions of project names.

### Propery references
The most significant improvement is related to property references. Previously there were three problems which prevented a lot of references to be replaced with their actual values:
1) Could not resolve substring references like `hello ${reference}!`
2) Could not resolve multiple references in one string (relates to 1) like `abc ${reference1} de ${reference2}`
3) Could not resolve references to DOM tree of `pom.xml` file like `${project.artifactId}`
This PR fixes all 3 above-mentioned problems, thus allowing for resolution of quite complex strings like `https://github.com/${project.groupId}/${project.artifactId}` to their actual value of for example`https://github.com/junit/junit` even if no properties are specified in the `pom.xml` file.

### Duplicate package names bug
POM Analyzer was saving package names as `<groupId>.<artifactId>` while OPAL renamed it `<groupId>:<artifactId>` and Metadata plugin was saving it as received from OPAL (with `:`), thus causing a lot of records in the database to be different when they are actually the same. For example, both `junit:junit` and `junit.junit` would be saved in the database while there should be only `junit:junit`.

### Extracting project names
The metadata database has a `packages.project_name` attribute which was always `null` because we did not have this information. In this PR POM Analyzer also extracts `project.name` (if it is present) from `pom.xml` and inserts it in the database along with producing it to Kafka topic.

## Testing
Changes were quite isolated thus were tested with unit testing using JUnit.

## Task list  
- [x] Resolution of multiple substring property references
- [x] Resolution of DOM tree property references
- [x] Change `.` to `:` when saving package name to the database
- [x] Extract and save project name

### DIFF
--- a/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
+++ b/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
@@ -116,6 +116,45 @@ public class DataExtractor {
     }
 
     /**
+     * Extracts project name from POM of certain Maven coordinate.
+     *
+     * @param groupId    groupId of the coordinate
+     * @param artifactId artifactId of the coordinate
+     * @param version    version of the coordinate
+     * @return Extracted project name as String
+     */
+    public String extractProjectName(String groupId, String artifactId, String version) {
+        String name = null;
+        try {
+            ByteArrayInputStream pomByteStream;
+            if ((groupId + ":" + artifactId + ":" + version).equals(this.mavenCoordinate)) {
+                pomByteStream = new ByteArrayInputStream(this.pomContents.getBytes());
+            } else {
+                pomByteStream = new ByteArrayInputStream(
+                        this.downloadPom(artifactId, groupId, version)
+                                .orElseThrow(FileNotFoundException::new).getBytes());
+            }
+            var pom = new SAXReader().read(pomByteStream).getRootElement();
+            if (this.resolutionMetadata == null || !this.resolutionMetadata.getLeft().equals(groupId + ":" + artifactId + ":" + version)) {
+                var metadata = this.extractDependencyResolutionMetadata(pom);
+                this.resolutionMetadata = new ImmutablePair<>(groupId + ":" + artifactId + ":" + version, metadata);
+            }
+            var properties = this.resolutionMetadata.getRight().getLeft();
+            var nameNode = pom.selectSingleNode("./*[local-name()='name']");
+            if (nameNode != null) {
+                name = replacePropertyReferences(nameNode.getText(), properties, pom);
+            }
+        } catch (DocumentException e) {
+            logger.error("Error parsing POM file for: "
+                    + groupId + ":" + artifactId + ":" + version);
+        } catch (FileNotFoundException e) {
+            logger.error("Error downloading POM file for: "
+                    + groupId + ":" + artifactId + ":" + version);
+        }
+        return name;
+    }
+
+    /**
      * Replaces all property references with their actual values.
      *
      * @param ref        String that can contain property references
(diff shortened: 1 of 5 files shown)


analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
```diff
@@ -115,6 +115,45 @@ public String extractPackagingType(String groupId, String artifactId, String ver
...
+        String name = null;
+        try {
+            ByteArrayInputStream pomByteStream;
+            if ((groupId + ":" + artifactId + ":" + version).equals(this.mavenCoordinate)) {
+                pomByteStream = new ByteArrayInputStream(this.pomContents.getBytes());
+            } else {
+                pomByteStream = new ByteArrayInputStream(
+                        this.downloadPom(artifactId, groupId, version)
+                                .orElseThrow(FileNotFoundException::new).getBytes());
+            }
```
ilyagrishkov, line 136: I would have replaced this part with a ternary operator. It looks cleaner and more readable. Same goes for the similar part in other `extract[...]`methods.

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
```diff
@@ -115,6 +115,45 @@ public String extractPackagingType(String groupId, String artifactId, String ver
...
+            var pom = new SAXReader().read(pomByteStream).getRootElement();
+            if (this.resolutionMetadata == null || !this.resolutionMetadata.getLeft().equals(groupId + ":" + artifactId + ":" + version)) {
+                var metadata = this.extractDependencyResolutionMetadata(pom);
+                this.resolutionMetadata = new ImmutablePair<>(groupId + ":" + artifactId + ":" + version, metadata);
+            }
+            var properties = this.resolutionMetadata.getRight().getLeft();
```
ilyagrishkov, line 142: IDE complains about duplicate code here. This part is present in many methods of the class. Is there a way to extract it into a separate method and reuse it?

analyzer/pom-analyzer/src/test/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractorTest.java
```diff
@@ -286,11 +286,14 @@ public void extractAllDataTest() {
         var actualSourcesUrl = dataExtractor.generateMavenSourcesLink("junit", "junit", "4.12");
         var expectedPackagingType = "jar";
         var actualPackagingType = dataExtractor.extractPackagingType("junit", "junit", "4.12");
+        var expectedProjectName = "JUnit";
```
ilyagrishkov, line 289: It might be worth testing a project where pom.xml doesn't contain the `name` attribute to cover both cases.

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
MihhailSokolov: Good point, I extracted those parts into separate methods

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
MihhailSokolov: Fixed that

analyzer/pom-analyzer/src/test/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractorTest.java
MihhailSokolov: Added test

PR #64, opened 2020-08-22
Fix NullPointerException in OPAL plugin

ilyagrishkov:
## Description
Fix NullPointerExeption when generating external CHA, when Instruction is not available. New changes replace absent information with "notFound"

## Motivation and context
After running the OPAL plugin on the entire Maven ecosystem and number of NullPointerException-s very thrown from the getCallSite() method. The cause of it was the absence of type and receiver information. 

## Testing
New unit tests are needed to test branches that include absent type and receiver information. 

## Task list  
- [x] Fix NPE when type and receiver information is not available.
- [x] Test the changes


MihhailSokolov (approved):
LGTM

PR #66, opened 2020-08-24
POM Analyzer support for multiple Maven repositories

MihhailSokolov:
## Description
This PR adds support for multiple Maven repositories to POM Analyzer. When POM Analyzer tries to download a `pom.xml` file from the first repository, if there is no such file, it will try the next repository and so on until the file is found and downloaded.
Maven repositories URLs are passed through environmental variable `MVN_REPO`, multiple URLs should be separated with `;`.

## Motivation and context
Previously we were for artifacts only on Maven Central repository, however there exist many more other repositories. So by including their URLs we can increase the success rate of resolving the artifacts.

## Testing
Testing was performed with JUnit using an artifact that is located at other repository than Maven Central. `org.tmatesoft.hg4j:hg4j:1.1.0-RELEASE` can only be found on `"https://maven.tmatesoft.com/content/repositories/releases/`. Therefore by checking that first we cannot resolve this artifact using only Maven Central, but when adding `tmatesoft` repository the artifact is properly resolved and analyzed.

## Task list  
- [X] Add support for multiple Maven repositories for POM Analyzer
- [X] When generating `sources.jar` link check that link is valid and points to the actual file


analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
```diff
@@ -31,12 +31,7 @@
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
ilyagrishkov, line 34: Checkstyle: avoid using "*" in imports. Default formatting can be fixed in IntelliJ code style settings so that Cmd+Alt+L doesn't create wild cards

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
```diff
@@ -71,8 +70,17 @@ public DataExtractor() {
      * @return Link to Maven sources jar file
      */
     public String generateMavenSourcesLink(String groupId, String artifactId, String version) {
-        return this.mavenRepos.get(0) + groupId.replace('.', '/') + "/" + artifactId + "/"
-                + version + "/" + artifactId + "-" + version + "-sources.jar";
+        for (var repo : this.mavenRepos) {
+            var url = repo + groupId.replace('.', '/') + "/" + artifactId + "/"
+                    + version + "/" + artifactId + "-" + version + "-sources.jar";
+            var file = httpGetToFile(url);
```
ilyagrishkov, line 76: In case of insufficient write permission httpGetToFile(url) will return Option.empty(), while `...-sources.jar` actually exists. Either check existence without downloading the file, or take a look at how error handling is done in OPAL's MavenCoordinate.downloadArtifact() method

ilyagrishkov (changes requested):

ilyagrishkov:
It's also worth increasing version number

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
MihhailSokolov: Fixed

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
MihhailSokolov: Done

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
```diff
@@ -511,11 +543,17 @@ public DependencyData extractDependencyData(String groupId, String artifactId, S
...
             var pomUrl = this.getPomUrl(artifactId, groupId, version, repo);
-            var pom = httpGetToFile(pomUrl).flatMap(DataExtractor::fileToString);
+            Optional<String> pom;
+            try {
+                pom = httpGetToFile(pomUrl).flatMap(DataExtractor::fileToString);
+            } catch (FileNotFoundException | UnknownHostException | MalformedURLException e) {
+                continue;
+            } catch (IOException e) {
+                throw new RuntimeException("Error downloading file: " + e.getMessage());
+            }
```
ilyagrishkov, line 556: Now when I think about it there might be an IO problem downloading from one repo, but we shouldn't immediately throw a RuntimeException, as there is a chance that it will be ok with other repos. I remember I suggested doing it this way during the conversation, but now when I see the code I think that this might be a problem.

ilyagrishkov (changes requested):

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/pom/DataExtractor.java
MihhailSokolov: How about now? In case of `IOException` the `httpGetToFile()` will return `Optional.empty()` and in the loop in `downloadPom()` it will just proceed to the next repo.

ilyagrishkov (approved):

PR #67, opened 2020-08-31
New callables insertion strategy

MihhailSokolov:
## Description
This PR changes the way that the callables of the new format CGs are inserted in the database in order to optimize the performance of insertion.
Previously,  all callables were inserted together and since a lot of external callables are already in the database, the performance of the PostgreSQL insertion was very low because conflicts on constraints caused a lot of deadlocks.
Now, internal callables are inserted with batch insert as before, then `select` is performed to retrieve IDs of the external callables that are already in the database and are in the CG, and then external callables which weren't in the database are inserted with the batch insert.

## Motivation and context
This PR should improve the performance of the metadata plugin.

## Testing
Used regression testing to make sure that overall functionality did not change and tested on a small CG to compare the results with the current metadata plugin on the `develop` branch.

## Task list  
- [X] Implement the new callables insertion strategy

### DIFF
--- a/analyzer/metadata-plugin/pom.xml
+++ b/analyzer/metadata-plugin/pom.xml
@@ -13,7 +13,7 @@
 
     <groupId>eu.fasten.analyzer</groupId>
     <artifactId>metadata-plugin</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
+    <version>0.1.1-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


MihhailSokolov:
@gousiosg Is this what you had in mind?

gousiosg:
LGTM

PR #68, opened 2020-09-03
QualityAnalyzer plugin

cg122:
## Description

QualityAnalyzer is FASTEN plugin to store/update code quality metadata from Kafka topic. There is a single Kafka message per single callable produced by [Rapid Plugin](https://github.com/fasten-project/quality-analyzer/tree/master/rapidplugin) using [Lizard](https://github.com/terryyin/lizard). Once a metadata message is consumed, the metadata is stored to respective database depending on the forge (mvn, c, python).

## Motivation and context

One of the use-case scenarios within FASTEN project is application level risk profile with external dependencies. The application-level risk considering a list of claimed transitive dependencies and this evaluation may cover code quality. Hence, a need to perform code quality analysis and to store the results.

## Testing the plugin

```
java -cp /path/to/projects/fasten/docker/server/server-0.0.1-SNAPSHOT-with-dependencies.jar eu.fasten.server.FastenServer -p /home/fasten/projects/fasten/docker/plugins/ -k localhost:9092 -pl "QualityAnalyzer" -kt "QualityAnalyzer=fasten.RapidPlugin.out" -d "mvn=jdbc:postgresql://postgres@localhost/dbname"
```
## Task list  

- [x] Create a plugin to consume topic produced by RapidPlugin
- [x] Create DB connection to the metadata DB
- [x]  Write query to do the DB updating for file/callable level metrics
- [x] Create test cases
- [x] Update documentation


analyzer/quality-analyzer/pom.xml
```diff
@@ -0,0 +1,161 @@
...
+            <artifactId>core</artifactId>
+            <version>0.0.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>eu.fasten</groupId>
+            <artifactId>server</artifactId>
+            <version>0.0.1-SNAPSHOT</version>
+            <!--            <scope>provided</scope>-->
+        </dependency>
+        <!--        NB! Uncomment the following 3 dependencies if you want to run Main class-->
```
MihhailSokolov, line 56: Please take a look at other plugins' pom files. We updated the dependency handling so now you don't need to comment/uncomment some dependencies to run the Main class

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+        private static DSLContext dslContext;
+        private final Logger logger = LoggerFactory.getLogger(QualityAnalyzer.class.getName());
+        private Throwable pluginError = null;
+        private String product = null;
+        private String artifact = null;
+        private String group = null;
+        private String version = null;
+        private JSONObject metrics = null;
+        private boolean restartTransaction = false;
+        private final int transactionRestartLimit = 3;
```
MihhailSokolov, line 43: `transactionRestartLimit` is now a part of Constants class in `core/data/Constants`, so please reference it from there

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+                    dslContext.transaction(transaction -> {
+                        metadataDao.setContext(DSL.using(transaction));
+                        long id;
+                        try {
+                            id = saveToDatabase(group + "." + artifact,
+                                    version,
+                                    metrics,
+                                    metadataDao);
+                        } catch (RuntimeException e) {
+                            logger.error("Error saving data to the database: '" + product + ":" + version + "'", e);
```
MihhailSokolov, line 94: `":"` is also one of the Constants. Please reference it as `Constants.mvnCoordinateSeparator` everywhere instead of hard-coding `":"`

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+                    });
+                } catch (Exception expected) {
+                }
+                transactionRestartCount++;
+            } while (restartTransaction && !processedRecord
+                    && transactionRestartCount < transactionRestartLimit);
+        }
+
+        public long saveToDatabase(String product, String version, JSONObject metrics, MetadataDao metadataDao) {
+            final var packageId = metadataDao.insertPackage(product, "mvn", null, null, null);
```
MihhailSokolov, line 120: `"mvn"` is also a Constant :)
`Constants.mvnForge`
In general if you have any hard-coded constants, please put them in the Constants class

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        @Override
+        public Optional<String> produce() {
+            return Optional.empty();
+        }
+
+        @Override
+        public String getOutputPath() {
+            return ".";
```
MihhailSokolov, line 135: Please add a proper output path like `groupId + "_" + artifactId + "_" + version + ".json"` You can take a look at what other plugins have as the output path.

MihhailSokolov (changes requested):
Looks great except some minor comments

analyzer/quality-analyzer/README.md
```diff
@@ -0,0 +1,46 @@
...
+- `-h` `--help` Show this help message and exit.
+- `-d` `--database` Comma separated key-value pairs of Database URLs for connection
+- `-f` `--file` Path to JSON file that contains JSON with Lizard tool generated quality metrics
+
+## Usage 
+
+To start as part of FASTEN server:
+
+```
+java -cp /path/to/projects/fasten/docker/server/server-0.0.1-SNAPSHOT-with-dependencies.jar eu.fasten.server.FastenServer -p /home/fasten/projects/fasten/docker/plugins/ -k localhost:9092 -pl "QualityAnalyzer" -kt "QualityAnalyzer=fasten.RapidPlugin.out" -d "java=jdbc:postgresql://postgres@localhost/dbname,python=jdbc:postgresql://postgres@localhost/pythondbname"
```
MihhailSokolov, line 29: I think you still need to provide a database password in the environmental variable, don't you?

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/Main.java
```diff
@@ -59,7 +62,8 @@ public static void main(String[] args) {
     public void run() {
         var metadataPlugin = new MetadataDatabasePlugin.MetadataDBExtension();
         try {
-            metadataPlugin.setDBConnection(PostgresConnector.getDSLContext(dbUrl, dbUser));
+            metadataPlugin.setDBConnection(new HashMap<>(Map.of("java",
```
MihhailSokolov, line 65: I think it would be a bit more convenient to use `forge` as a key in the map of DSL Contexts instead of language. Because during runtime you choose the right context based on the forge you receive, not language. But if you prefer having a language there, I don't really have a problem with that.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/MetadataUtils.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /**
+     *
+     * @param forge         String which could have value MVN, PyPI or C.
+     * @param jsonRecord    Object that contains quality analysis metadata.
+     */
+    public void insertMetadataIntoDB(String forge, JSONObject jsonRecord) {
+
+        if(forge.equalsIgnoreCase(QAConstants.MVN_FORGE)) {
```
MihhailSokolov, line 68: We already have a constants class which you can use. It's in the `core/data/Constants` and also have `mvnForge`.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/MetadataUtils.java
```diff
@@ -0,0 +1,273 @@
...
+                .from(PackageVersions.PACKAGE_VERSIONS)
+                .where(PackageVersions.PACKAGE_VERSIONS.PACKAGE_ID.equal(pkgId))
+                .and(PackageVersions.PACKAGE_VERSIONS.VERSION.equal(version))
+                .fetchOne();
+
+        if(pkgVersionRecord != null) {
+            return pkgVersionRecord.getId();
+        }
+
+        return -1L;
```
MihhailSokolov, line 193: Package version with ID = -1 is also already taken for external callable reference. Same goes for modules and packages.
```sql
fasten_java=> select * from packages where id = -1;
 -1 | external_callables_library | mvn   |              |            | 

fasten_java=> select * from package_versions where id = -1;
 -1 |         -1 | 0.0.1   | OPAL         |            |          | 

fasten_java=> select * from modules where id=-1;
 -1 |                 -1 | global_external_callables |            | 
```

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+    @Extension
+    public static class QualityAnalyzer implements KafkaPlugin, DBConnector {
+
+        private final Logger logger = LoggerFactory.getLogger(QualityAnalyzer.class.getName());
+        private String consumerTopic = "fasten.RapidPlugin.out";
+        private MetadataUtils utils = null;
+
+        @Override
+        public void setDBConnection(Map<String, DSLContext> dslContexts) {
+            this.utils = new MetadataUtils(dslContexts);
```
MihhailSokolov, line 57: Please make `utils` static and here use `QualityAnalyzer.utils = new MetadataUtils(dslContexts)`. Otherwise you will get a `NullPointerException` when running this plugin with FASTEN Server

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+
+            if (jsonRecord.has("payload")) {
+                forge = jsonRecord
+                        .getJSONObject("payload")
+                        .getString("forge".replaceAll("[\\n\\t ]", ""));
+            }
+
+            logger.info("forge = " + forge);
+
+            //TODO: what if forge = null? Throw an exception?
```
MihhailSokolov, line 86: No need. The way you have it with that if-else it just find

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+        public void start() {
+        }
+
+        @Override
+        public void stop() {
+        }
+
+        @Override
+        public Throwable getPluginError() {
+            return null;
```
MihhailSokolov, line 131: You should add a field something like `pluginError` which you return here. Then wrap everything in `consume` method in try-catch block which catches all `RuntimeException`, logs them, and sets that `pluginError` equal to the thrown exception.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+                        .getJSONObject("payload")
+                        .getString("forge".replaceAll("[\\n\\t ]", ""));
+            }
+
+            logger.info("forge = " + forge);
+
+            //TODO: what if forge = null? Throw an exception?
+
+            if (forge != null) {
+                utils.insertMetadataIntoDB(forge, jsonRecord);
```
MihhailSokolov, line 89: Also, I think that all inserts should be done within one transaction because otherwise, we might end up with partial data.
You can see how transaction management with retries works [here](https://github.com/fasten-project/fasten/blob/develop/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java#L140)

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/data/QAConstants.java
```diff
@@ -0,0 +1,38 @@
...
+
+package eu.fasten.analyzer.qualityanalyzer.data;
+
+public class QAConstants {
+
+    public static final String QA_VERSION_NUMBER = "0.0.1";
+    public static final String QA_PLUGIN_NAME = "Quality Analyzer Plugin";
+
+    //needs to be integrated with core.data.Constants
+    public static final String MVN_FORGE = "mvn";
```
MihhailSokolov, line 27: I think that these 3 forges should be placed in that `Constants` class we have in the `core/data` because they will be used by multiple plugins. And Maven forge is already there.
Other constants can be here as they are.

MihhailSokolov (changes requested):
Overall, looks good. Nice work! However, I have some comments for improvements :)

analyzer/quality-analyzer/README.md
zikanl: Yes, that is correct, will update the documentation.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/MetadataUtils.java
zikanl: Correct, I have already updated QAConstants and synched with Constants. However, I wait for the final decision about the forges, before updating it altogether. BTW, Constants do not have any reference to either C or Python forge.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/Main.java
zikanl: I didn't implement this piece of code, simply merged multiple-db-conc, before finalizing QA code.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
zikanl: OK, done.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
zikanl: OK, removed TODO

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/data/QAConstants.java
zikanl: I agree, the comment says it all (`//needs to be integrated with core.data.Constants`)

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
zikanl: OK, done

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/MetadataUtils.java
zikanl: OK, done.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/data/QAConstants.java
zikanl: OK, done

analyzer/quality-analyzer/README.md
zikanl: OK, done

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/MetadataUtils.java
zikanl: OK, done. Null is returned instead of -1L, and checked on it.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
zikanl: OK, done

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
zikanl: I followed closely the implementation of the VulnerabilityConsumer - I do only updates of the metadata in the callables (using updateCallableMetadata from MetadataDao). Therefore, I have changed the function names to better illustrate this.

MihhailSokolov: As long as you are certain that we won't be left with partial data in the case when your plugin unexpectedly terminates in the middle of processing, I am fine with what you have. But if there is a chance that part of the data will be written and the other part won't, then proper transaction management should be implemented.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+            this.pluginError = null;
+
+            if (forge != null) {
+                try {
+                    utils.updateMetadataInDB(forge, jsonRecord);
+                } catch(Exception ex) {
+                    setPluginError(ex);
+                }
+            } else {
+                logger.error("Could not extract forge from the message");
```
MihhailSokolov, line 97: Please here add something like `this.pluginError = new RuntimeError("Could not extract forge from the message");`. Otherwise, when there is no forge, it will skip the record, and it will seem that it was processed successfully (no errors) while in reality, it wasn't successful at all and there was an error (missing forge).

zikanl: OK done, with a little bit refactoring and setting pluginError to RuntimeException().

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
zikanl: OK, done, I refactored the suggested code. BTW, the code from MetadataDatabasePlugin should be refactored and simplified IMO.

MihhailSokolov: Thanks. Yes, it is already being refactored by Stefanos (AUEB) to allow the processing of C and Python callgraphs.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -0,0 +1,196 @@
...
+                return;
+            }
+
+            boolean processedRecord = false;
+            int transactionRestartCount = 0;
+            boolean restartTransaction = false;
+
+            Long recordId = null;
+
+            do {
```
MihhailSokolov, line 103: If you are trying to use transactions, then you are not actually starting a transaction. Here is the link that nicely explains how JOOQ transaction management works https://www.jooq.org/doc/3.12/manual/sql-execution/transaction-management/
If you don't want transactions but only retries, then you don't need that empty catch block. It is needed for proper error management in transactions (you can take a closer look at how this works in the `MetadataDatabasePlugin` class - transaction starts with lambda function, there is an inner catch block which decides whether to restart transaction or not and it always throws caught exception back from the transaction lambda to stop the transaction, then outer empty catch block catches everything but doesn't do anything because everything was already done in the inner catch block, and then transaction is restarted if needed).

zikanl: OK, the latest version is there. I moved transactions to `MetadataUtils`, in order (hopefully) to get a more readable code, and hopefully I handle all Exceptions. I still need to throw `JsonParsingException` instead of `IllegalStateException` where applicable.

MihhailSokolov: Yes, the transaction management looks correct now. Thanks!

MihhailSokolov (approved):
Everything looks good to me, nice work. If this is completely finished, @zikanl you can merge it.

PR #70, opened 2020-09-08
Refactor constants

MihhailSokolov:
## Description
This PR creates a new class in `core/data` called `Constants` which contains several constants used throughout the project.

## Motivation and context
It's a good practice to keep the constants in one place instead of hard-coding them in several places.

## Testing
Regression testing - ran existing tests to make sure everything works as it was.

## Task list  
- [x] Create `Constants` class and put all hard-coded constants there

### DIFF
--- a/core/src/main/java/eu/fasten/core/data/Constants.java
+++ b/core/src/main/java/eu/fasten/core/data/Constants.java
@@ -2,9 +2,9 @@ package eu.fasten.core.data;
 
 public class Constants {
 
-    public static final String mvnCoordinatePartsJoin = ":";
+    public static final String mvnCoordinateSeparator = ":";
 
-    public static final String mavenForge = "mvn";
+    public static final String mvnForge = "mvn";
 
     public static final String opalGenerator = "OPAL";
 
@@ -13,5 +13,9 @@ public class Constants {
     public static final int insertionBatchSize = 4096;
 
     public static final String mvnRepoEnvVariable = "MVN_REPO";
+
+    public static final String fastenDbPassEnvVariable = "FASTEN_DBPASS";
+
+    public static final String pgPasswordEnvVariable = "PGPASSWORD";
 
 }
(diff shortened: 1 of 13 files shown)


core/src/main/java/eu/fasten/core/data/Constants.java
```diff
@@ -0,0 +1,17 @@
+package eu.fasten.core.data;
+
+public class Constants {
+
+    public static final String mvnCoordinatePartsJoin = ":";
```
ilyagrishkov, line 5: I would call it mvnCoordinateSeparator

core/src/main/java/eu/fasten/core/data/Constants.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public static final String mavenForge = "mvn";
```
ilyagrishkov, line 7: You use `mvn` in other cases, but `maven` here. I think this constant should be called `mvnForge`

core/src/main/java/eu/fasten/core/data/Constants.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public static final String opalGenerator = "OPAL";
+
+    public static final int transactionRestartLimit = 3;
+
+    public static final int insertionBatchSize = 4096;
+
+    public static final String mvnRepoEnvVariable = "MVN_REPO";
```
ilyagrishkov, line 15: Maybe also add FASTEN_DB_PASS env variable constant?

ilyagrishkov (changes requested):

core/src/main/java/eu/fasten/core/data/Constants.java
MihhailSokolov: Renamed

core/src/main/java/eu/fasten/core/data/Constants.java
MihhailSokolov: Renamed it as well

core/src/main/java/eu/fasten/core/data/Constants.java
MihhailSokolov: Sure, added it along with `PGPASSWORD`

ilyagrishkov (approved):
LGTM

Issue #89, opened 2020-10-20
KnowledgeBase/REST API integration tests


marcomicera:
- [x] PostgreSQL integration tests (#92)
- [ ] ~~RESTeasy~~ Tomcat integration tests (changed in #123)

marcomicera:
On hold, as #87 might cause the Servlet engine to change.

PR #80, opened 2020-09-29
Final KB REST API

marcomicera:
## Tasks

- [x] [Endpoints table and their progress](https://github.com/fasten-project/fasten/wiki/API-endpoints-for-Maven-projects)
- ~~API specifications~~ (dropped in d13d671a, response data structures already generated by JOOQ)
- [x] Server stubs 0c9eea82
- [x] Queries
- [ ] Tests
- [x] Pagination (#88, d1b7acba0581acc0efe261bb4149ddbd8f49ef49)

PR #92, opened 2020-10-21
Added PostgresConnector tests

marcomicera:
Part of the tests (#89) of the REST API (#80) ended up being tests for the [PostgresConnector](https://github.com/fasten-project/fasten/blob/develop/server/src/main/java/eu/fasten/server/connectors/PostgresConnector.java) class.

Here I'm using [Testcontainers](https://www.testcontainers.org) to fire up a PostgreSQL container and trying to connect to it through that connector class.

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -22,6 +22,20 @@
             <artifactId>core</artifactId>
             <version>0.0.1-SNAPSHOT</version>
         </dependency>
+
+        <!-- Integration tests -->
+        <dependency>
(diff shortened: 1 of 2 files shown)


MihhailSokolov (approved):
Looks good to me, thanks!

MihhailSokolov:
@gousiosg should we merge this?

gousiosg:
I believe so; since those dependencies are test only. Do they work in our CI? If yes, let's merge.

PR #96, opened 2020-10-22
Add GitHub Super Linter

koppor:
## Description

GitHub introduced GitHub Super Linter: one linter to rule them all. See <https://github.blog/2020-06-18-introducing-github-super-linter-one-linter-to-rule-them-all/> for details. Since I am a huge fan of linting (especially [markdown-lint](https://github.com/igorshubovych/markdownlint-cli#readme)) and I was seeing that "checkstyle" is already in place, I thought, this could be helpful.

## Motivation and context

Having linting issues reported in a clear way increases the likelihood that linting issues are addressed during submission of a pull request.

I did not touch the maven build file since I am not sure whether you really like this new trendy GitHub tool.

Alternatively, it is also possible that the PR is directly commented ([reviewdog](https://github.com/reviewdog/reviewdog)), but this causes much noise at unfinished PR. The alternative to that is [PR annotation](https://github.com/staabm/annotate-pull-request-from-checkstyle), but this fails if an external contributor is contributing something due to token issues. Thus, I tink, the GitHub super linter is a good thing.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,24 @@
+name: Lint Code Base
+# Documentation:
+# https://help.github.com/en/articles/workflow-syntax-for-github-actions
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    name: Lint Code Base
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          # Full git history is needed to get a proper list of changed files within `super-linter`
+          fetch-depth: 0
+      - name: Lint Code Base
+        uses: github/super-linter@v3
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          DEFAULT_BRANCH: develop
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/linter.yml
```diff
@@ -0,0 +1,55 @@
...
+
+#############################
+# Start the job on all push #
+#############################
+on:
+  push:
+    branches-ignore: [master]
+    # Remove the line above to run when pushing to master
+  pull_request:
+    branches: [master]
```
amottier, line 22: As most of the PR are currently made on `develop` branch I guess that we should use `develop` here and not `master`

.github/workflows/linter.yml
```diff
@@ -0,0 +1,55 @@
...
+          fetch-depth: 0
+
+      ################################
+      # Run Linter against code base #
+      ################################
+      - name: Lint Code Base
+        uses: github/super-linter@v3
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          DEFAULT_BRANCH: master
```
amottier, line 54: Same as previous comment, I think it would make sense to target `develop` branch.

amottier (commented):
I'm in favor of adding this new tool to experiment. We will see if it's really help us or bring to much noise.

@koppor can you read my comment and let me know what you think?

proksch:
This PR is open for almost a year now... do we still want to integrate this change?

koppor:
@proksch Could you please rebase my PR to `develop`?

I included the requested changes (was replacing `master` by `develop`)

amottier:
@koppor I changed the targeted branch from `master` to `develop` and merged your change.

@proksch I think we can experiment GitHub Super Linter. If we find it to "noisy" we can still revert this change easily as it is located in a single file (`.github/workflows/linter.yml`).

mir-am:
Thanks for the contribution.
I do appreciate code linters. However, GH super linter is aggressive and it generates many warnings for the project.
Currently, I would like to remove the linter from our GH workflow. Later on, we can configure a linter and fix its warnings in a separate PR.
That being said, we also have `checkstyle` in the project, which can be configured and run in the CI.

proksch:
While I generally do not oppose the idea to have a linter in the project, it seems that we would need to adopt its configuration to the project before we can use it, as the defaults are often rather meaningless (80 char line length, seriously?). The linter is of no help to anybody, if it blocks further development until we have solved 100s of errors.

PR #97, opened 2020-10-23
Support for multiple DB connections in the FASTEN server

mir-am:
## Description
FASTEN server now accepts a set of DB URLs for creating multiple DB connections. For example, to connect to Java and Python DBs, the argument `-d` in the server will be set as follows:
```
-d java=jdbc:postgresql://postgres@localhost/dbname,py=jdbc:postgresql://postgres@localhost/dbname
```
In the plug-in, by using `dslContexts.get("java")`, the connection to the Java DB can be used.
I propose `java`, `py` and `c` for the key names of the list of DBs in CLI and in the code.

## Motivation and context
Plug-ins like QualityAnlyzer and REST API require multiple DB connections as they need to access metadata for different languages.

## Testing
I locally tested the FASTEN server to make sure that it successfully create a DB connection for metadata plug-in and POMAnalyzer. However, we still need to test both plug-ins on our servers to ensure that they work as expected.

## Task list  
- [x] Add the support for multiple DB connections in the FASTEN server
- [x] Change `DBConnector` interface to for accepting a map of DB connections.
- [x] Make the plug-ins that require a DB connection compatible with the new changes.

## Additional context
None

### DIFF
--- a/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/Main.java
+++ b/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/Main.java
@@ -18,6 +18,7 @@
 
 package eu.fasten.analyzer.metadataplugin;
 
+import eu.fasten.core.data.Constants;
 import eu.fasten.server.connectors.PostgresConnector;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
@@ -62,7 +63,7 @@ public class Main implements Runnable {
     public void run() {
         var metadataPlugin = new MetadataDatabasePlugin.MetadataDBExtension();
         try {
-            metadataPlugin.setDBConnection(new HashMap<>(Map.of("java",
+            metadataPlugin.setDBConnection(new HashMap<>(Map.of(Constants.mvnForge,
                     PostgresConnector.getDSLContext(dbUrl, dbUser))));
         } catch (IllegalArgumentException | SQLException e) {
             logger.error("Could not connect to the database", e);
--- a/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/Main.java
+++ b/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/Main.java
@@ -22,6 +22,9 @@ import eu.fasten.server.connectors.PostgresConnector;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+
 import org.json.JSONObject;
 import org.json.JSONTokener;
 import org.slf4j.Logger;
@@ -59,7 +62,8 @@ public class Main implements Runnable {
     public void run() {
         var metadataPlugin = new MetadataDatabasePlugin.MetadataDBExtension();
         try {
-            metadataPlugin.setDBConnection(PostgresConnector.getDSLContext(dbUrl, dbUser));
+            metadataPlugin.setDBConnection(new HashMap<>(Map.of("java",
+                    PostgresConnector.getDSLContext(dbUrl, dbUser))));
         } catch (IllegalArgumentException | SQLException e) {
             logger.error("Could not connect to the database", e);
             return;
--- a/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java
+++ b/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePlugin.java
@@ -76,7 +76,7 @@ public class MetadataDatabasePlugin extends Plugin {
 
         @Override
         public void setDBConnection(Map<String, DSLContext> dslContexts) {
-            MetadataDBExtension.dslContext = dslContexts.get("java");
+            MetadataDBExtension.dslContext = dslContexts.get(Constants.mvnForge);
         }
 
         @Override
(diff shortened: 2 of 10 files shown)


MihhailSokolov (approved):
Everything looks good, thanks!
Only I would suggest using forge as a key instead of language because during runtime forges will be used to select the right DSL Context.
Also, do I understand correctly that these changes assume that all databases have the password? This is not urgent, but in the future, we could also expand this functionality to support different passwords for each database.

mir-am:
> Everything looks good, thanks!
> Only I would suggest using forge as a key instead of language because during runtime forges will be used to select the right DSL Context.

what is the name of forges for the three languages?

> Also, do I understand correctly that these changes assume that all databases have the password? This is not urgent, but in the future, we could also expand this functionality to support different passwords for each database.

Yes. It uses the same password for all DBs. However, in the future, we can support a different password for each DB.

MihhailSokolov:
> what is the name of forges for the three languages?

Well, for Maven it is `mvn` as you probably know. Constants like forges must be stored in the `Constants` class in `core/data`. `mvnForge` is already there.
But I am not sure about the other 2. I thought it should be `pypi` and `debian`, however in QualityAnalyzer they are just called `python` and `C`.
@StefanosChaliasos and @zikanl We should agree on a single representation for Python and C forges, place it in `core/data/Constants` and use the same representation everywhere.

> Yes. It uses the same password for all DBs. However, in the future, we can support a different password for each DB.

Yes, I suppose we should :)

zikanl:
> > what is the name of forges for the three languages?
> 
> Well, for Maven it is `mvn` as you probably know. Constants like forges must be stored in the `Constants` class in `core/data`. `mvnForge` is already there.
> But I am not sure about the other 2. I thought it should be `pypi` and `debian`, however in QualityAnalyzer they are just called `python` and `C`.
> @StefanosChaliasos and @zikanl We should agree on a single representation for Python and C forges, place it in `core/data/Constants` and use the same representation everywhere.

I agree on this one, we should have two constants defined in core.data.Constants for python and C forges. The only reason why QualityAnalyzer uses python and C was that @mir-am proposed this. 

Who is going to update the core.data.Constants then? Also, we need to update all Plugins and Fasten Server that contain "java" as hard-coded key (and replace it by Constants.mvnForge)

MihhailSokolov:
> > > what is the name of forges for the three languages?
> > 
> > 
> > Well, for Maven it is `mvn` as you probably know. Constants like forges must be stored in the `Constants` class in `core/data`. `mvnForge` is already there.
> > But I am not sure about the other 2. I thought it should be `pypi` and `debian`, however in QualityAnalyzer they are just called `python` and `C`.
> > @StefanosChaliasos and @zikanl We should agree on a single representation for Python and C forges, place it in `core/data/Constants` and use the same representation everywhere.
> 
> I agree on this one, we should have two constants defined in core.data.Constants for python and C forges. The only reason why QualityAnalyzer uses python and C was that @mir-am proposed this.
> 
> Who is going to update the core.data.Constants then? Also, we need to update all Plugins and Fasten Server that contain "java" as hard-coded key (and replace it by Constants.mvnForge)

I think it should be `pypi` and `debian`. I can add them to `Constants`, but I want to first hear what @StefanosChaliasos thinks because he is the one implementing these languages in our system.

StefanosChaliasos:
> > > > what is the name of forges for the three languages?
> > > 
> > > 
> > > Well, for Maven it is `mvn` as you probably know. Constants like forges must be stored in the `Constants` class in `core/data`. `mvnForge` is already there.
> > > But I am not sure about the other 2. I thought it should be `pypi` and `debian`, however in QualityAnalyzer they are just called `python` and `C`.
> > > @StefanosChaliasos and @zikanl We should agree on a single representation for Python and C forges, place it in `core/data/Constants` and use the same representation everywhere.
> > 
> > 
> > I agree on this one, we should have two constants defined in core.data.Constants for python and C forges. The only reason why QualityAnalyzer uses python and C was that @mir-am proposed this.
> > Who is going to update the core.data.Constants then? Also, we need to update all Plugins and Fasten Server that contain "java" as hard-coded key (and replace it by Constants.mvnForge)
> 
> I think it should be `pypi` and `debian`. I can add them to `Constants`, but I want to first hear what @StefanosChaliasos thinks because he is the one implementing these languages in our system.

I think that we should use `pypi` and `debian` because we are referring to forges. The problem here is that in some plug-ins, such as MetadataPlugin, we use forge to decide which implementation to use. Maybe is a good idea to add another field called language? For example, what will happen if we choose to analyze projects from Github? Another example would be to add support for a language that uses Maven or support another C ecosystem. In general, we have different implementations for languages, but sometimes we selected them based on the forge. In the future, we may also need to provide different implementations for forges—for instance, mvn-java vs mvn-kotlin, or mvn-java vs github-java, etc.

MihhailSokolov:
Aright, then tomorrow I will add `debian` and `pypi` to `Constants` and update this PR accordingly. Then it can be merged. And then @StefanosChaliasos and @zikanl should adjust their code accordingly.

MihhailSokolov:
> Aright, then tomorrow I will add `debian` and `pypi` to `Constants` and update this PR accordingly. Then it can be merged. And then @StefanosChaliasos and @zikanl should adjust their code accordingly.

It's done. @StefanosChaliasos and @zikanl please update your code to use the `Constants` class.
@mir-am could you take a look at my changes? And everything is fine please merge the PR. Thanks!

StefanosChaliasos:
> > Aright, then tomorrow I will add `debian` and `pypi` to `Constants` and update this PR accordingly. Then it can be merged. And then @StefanosChaliasos and @zikanl should adjust their code accordingly.
> 
> It's done. @StefanosChaliasos and @zikanl please update your code to use the `Constants` class.
> @mir-am could you take a look at my changes? And everything is fine please merge the PR. Thanks!

Done for the metadata plugin

PR #99, opened 2020-10-27
Added Endocode contributors

marcomicera:
I've used the [`contributors` tag](https://maven.apache.org/pom.html#contributors), feel free to move it elsewhere if needed.


mir-am:
It's okay to merge. 
@ilyagrishkov. we should do the same for the TUD team and add our developers to the main POM file.

PR #91, opened 2020-10-21
Add support for C

StefanosChaliasos:
## Description
Refactoring of `ExtendedRevisionCallGraph.java` and `MetadataDabasePlugin.java` to make it easier to support more languages. Specifically, regarding `ExtendedRevisionCallGraph` I have created the abstract classes `ExtendedRevisionCallGraph.java`, `Node.java`, `ExtendedBuilder.java`. Moreover, I provide implementations for C and Java. For `MetadataDabasePlugin`, I created a general class `MetadataDBExtension` and two sub-classes `MetadataDBCExtension` and `MetadataDBJavaExtension`.
To add support for a new language, implementations for `ExtendedRevisionCallGraph`, `Node`, `ExtendedBuilder`, and a sub-class for `MetadataDabasePlugin` are needed.

## Motivation and context
The code was extremely coupled with Java's call graphs.

## Testing
* Refactoring of old tests.
* ExtendedRevisionCCallGraphTest.java
* Manually tested MetadataDBCExtension via its Main.java.

## Task list  
- [x] MetadataDBCExtension and MetadataDBJavaExtension cannot be load by the FASTEN server.
- [x] TODO ExtendedBuilder.java (Line 35)
- [x] FIXME MetadataDatabasePlugin.java (Line 196)
- [x] Add tests for MetadataDatabaseCPlugin.java

PR #107, opened 2020-11-06
Python metadata plugin

vitsalis:
## Description
Introduce metadata plugin for inserting Python call graphs into the metadata database.
This pull request is based on #91 , and therefore it should be merged after it.

## Motivation and context
This is Python specific code which is very similar to the ones implemented for Java and C.

## Testing
1. Unit tests
2. Local tests on my machine (OS x) using a kafka topic which stored either the call graphs themselves or a path to the call graph.
3. Repeated the same tests on the monster server.

## Additional context
Based on #91 .

### DIFF
--- a/analyzer/metadata-plugin/pom.xml
+++ b/analyzer/metadata-plugin/pom.xml
@@ -85,6 +85,19 @@
                                         <Plugin-License>Apache License 2.0</Plugin-License>
                                     </manifestEntries>
                                 </transformer>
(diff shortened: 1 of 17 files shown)


StefanosChaliasos:
It also fixes a bug https://github.com/fasten-project/fasten/pull/107/commits/21a632282ad6ae0848bd930916a952e6fac352b2 related to the fasten server loading metadata plug-in

PR #103, opened 2020-11-03
Python metadata plugin

vitsalis:
## Description
Introduce metadata plugin for inserting Python call graphs into the metadata database.
This pull request is based on #91 , and therefore it should be merged after it.

## Motivation and context
This is Python specific code which is very similar to the ones implemented for Java and C.

## Testing
1. Unit tests
2. Local tests on my machine (OS x) using a kafka topic which stored either the call graphs themselves or a path to the call graph.
3. Repeated the same tests on the monster server.

## Additional context
Based on #91 .

PR #109, opened 2020-11-06
Add Python Metadata Plug-in

StefanosChaliasos:
## Description
See #103


core/src/main/java/eu/fasten/core/data/Constants.java
```diff
@@ -12,6 +12,8 @@
 
     public static final String debianForge = "debian";
 
+    public static final String pypiForge = "PyPI";
```
StefanosChaliasos, line 15: @vitsalis Could you please keep one of pypiForge or pyForge, and update the code base accordingly?

PR #110, opened 2020-11-06
Fix bug in makeDBConnection

StefanosChaliasos:
The stream was producing a map that had a string composed of forge=uri
as a key. In the map we want to have forge as the key to be able to
select the connection based on the forge.

@mir-am @MihhailSokolov Please verify that all the plug-ins that use the databases have the same logic to connect to the databases.

### DIFF
--- a/server/src/main/java/eu/fasten/server/FastenServer.java
+++ b/server/src/main/java/eu/fasten/server/FastenServer.java
@@ -231,7 +231,7 @@ public class FastenServer implements Runnable {
         dbPlugins.forEach((p) -> {
             if (dbUrls != null) {
                 p.setDBConnection(dbUrls.entrySet().stream().map(e -> new AbstractMap.SimpleEntry<>(e.getKey(),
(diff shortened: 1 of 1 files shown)


mir-am:
> The stream was producing a map that had a string composed of forge=uri
> as a key. In the map we want to have forge as the key to be able to
> select the connection based on the forge.
> 
> @mir-am @MihhailSokolov Please verify that all the plug-ins that use the databases have the same logic to connect to the databases.

Thanks for finding and fixing the bug. Indeed, there was a null pointer exception because of the buggy map in `dslContexts.get(Constants.mvnForge)`. A simple check like the below example reveals the bug.
```
MetadataDBJavaExtension.dslContext.createTemporaryTable("test");
```
It now works fine with your fix. Other plug-ins also use the same method to connect, i.e. `dslContexts.get(Constants.mvnForg);`, hence they should also work fine.

StefanosChaliasos:
@mir-am great! Could you please add a test?

PR #115, opened 2020-11-14
Start implementing JsonUtils for ercg

ashkboos:
## Description
This PR main objective is to decrease the amount of memory that the opal plug-in requires for processing Maven. Since we have already improved the memory usage of the intermediate code, here we only focus on improving the serialization.

## Motivation and context
For big packages sometimes we face OOM while serializing the CGs. Since we convert RCGs to JsonObject before writing it to String, there is room for improvement by removing JsonObject and converting directly to String. 

## Testing
We implement unit tests that check whether the serialized strings are completely identical to what we used to produce previously. We also compare the deserialized version of the strings with the previous implementation to make sure that ercg objects are equal.

## Task list  
- [x] Create a JSONUtil class. This class goal is to gather json serialization and deserialization aspects of data libraries in the core.
- [x] Implement toJson for Java ERCG
- [x] Test toJson by comparing Strings to the previous implementation String
- [x] Test toJson by comparing the deserialized objects

## Additional context
Note that our experiments showed that using serialization libraries can not help and make the situation even worse.


mir-am:
@ashkboos 
What is the current status of this PR?
Can someone review it for merging it?

analyzer/javacg-opal/pom.xml
```diff
@@ -45,6 +45,11 @@
             <version>3.1.4</version>
             <type>pom</type>
         </dependency>
+        <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.0</version>
```
amottier, line 51: I would suggest to add `<scope>test</scope>` for `jsonassert` dependency.

core/src/main/java/eu/fasten/core/data/ExtendedRevisionJavaCallGraph.java
```diff
@@ -169,4 +173,63 @@ public String getRevisionName() {
         return artifactId + "_" + groupId + "_" + this.version;
     }
 
+    @Override
```
amottier, line 176: Maybe implementing `equals` and `hashCode` in parent class that actually declare the field `nodeCount`, `graph` would make more sense.

core/src/main/java/eu/fasten/core/data/Graph.java
```diff
@@ -195,4 +195,35 @@ public JSONObject toJSON() {
         result.put("resolvedCalls", resolvedCallsJSON);
         return result;
     }
+
```
amottier, line 198: LGTM

core/src/main/java/eu/fasten/core/data/JavaNode.java
```diff
@@ -62,4 +62,28 @@ public FastenURI changeName(final String className, final String methodName) {
         final var uri = this.getUri().toString().replace("/" + getClassName() + ".", "/" + className + ".");
         return FastenURI.create(uri.replace("." + getMethodName() + "(", "." + methodName + "("));
     }
+
```
amottier, line 65: I have the same kind of comment as with `ExtendedRevisionJavaCallGraph`

amottier:
@ashkboos, @mir-am if this PR is ready for review I can review it (actually I already started with a few minor feedback).

analyzer/javacg-opal/pom.xml
ashkboos: Thanks for your comment I addressed it.

core/src/main/java/eu/fasten/core/data/ExtendedRevisionJavaCallGraph.java
ashkboos: Thanks for the comment, I agree, In fact, all of the data classes must have equals and hashCode. In this PR I had to implement a java specific one to do the testing but we will have another PR in the future for refactoring the whole data package that needs to address all these.

core/src/main/java/eu/fasten/core/data/JavaNode.java
ashkboos: The same answer also applies here.

core/src/main/java/eu/fasten/core/data/JSONUtils.java
```diff
@@ -0,0 +1,287 @@
...
+            } else {
+                result.append("\"").append(entry.getKey()).append("\":").append(entry.getValue())
+                    .append(",");
+            }
+        }
+        removeLastIfNotEmpty(result, map.size());
+    }
+
+    /**
+     * Qoutes a given String.
```
amottier, line 223: Typo in the javadoc and in function name

core/src/main/java/eu/fasten/core/data/JSONUtils.java
```diff
@@ -0,0 +1,287 @@
...
+        return result.toString();
+    }
+
+    /**
+     * Appends general information of the revision to the beginning of the StringBuilder.
+     *
+     * @param erjcg  the object to extract the information from.
+     * @param result the StringBuilder to append the information.
+     */
+    private static void appendArtifactInformation(ExtendedRevisionJavaCallGraph erjcg,
```
amottier, line 34: I would suggest to have the `StringBuilder` as the first parameter for coherency with other methods.

core/src/main/java/eu/fasten/core/data/JSONUtils.java
```diff
@@ -0,0 +1,287 @@
...
+     *
+     * @param erjcg and object of java revision call graph to be converted to JSON String.
+     * @return the corresponding JSON String.
+     */
+    public static String toJSONString(final ExtendedRevisionJavaCallGraph erjcg) {
+        var result = new StringBuilder("{");
+        appendArtifactInformation(erjcg, result);
+        appendCha(result, erjcg.classHierarchy);
+        appendGraph(result, erjcg.graph);
+        if (erjcg.timestamp >= 0) {
```
amottier, line 21: In case timestamp value is negative (probably not a valid scenario) we will have a trailing comma coming from `appendGraph`call.

JSON Java accept the JSON with the trailing comma whereas JSON specification don't (see https://github.com/stleary/JSON-java/issues/269#issuecomment-239738561) so I think this is not a real issue.

I'm reporting this because we don't get strictly the same result as with the previous implementation.

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/data/JSONUtilsTest.java
```diff
@@ -0,0 +1,157 @@
+package eu.fasten.analyzer.javacgopal.data;
```
amottier, line 1: I'm curious why the test class is not in the same package as the tested class (`eu.fasten.core.data`)?

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/data/JSONUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
+            readDataCSV(new File(Objects.requireNonNull(Thread.currentThread().getContextClassLoader()
+            .getResource("121Coordinates.csv")).getFile()));
+    }
+
+    @Test
+    void toJSONString() throws IOException {
+
+        final var ser1 = avgConsumption(graph, "direct", "direct", 20, 20);
+        final var ser2 = avgConsumption(graph, "jsonObject", "jsonObject", 20, 20);
+        JSONAssert.assertEquals(ser1, ser2, JSONCompareMode.LENIENT);
```
amottier, line 62: I would rather use `JSONCompareMode.STRICT` or `JSONCompareMode.NON_EXTENSIBLE` because we want to have exactly the same JSON fields with the two implementation (see http://jsonassert.skyscreamer.org/apidocs/org/skyscreamer/jsonassert/JSONCompareMode.html)

amottier (commented):
Overall I think I have only minor comments. @ashkboos maybe I let you answer to my comments and then we can decide if we merge the PR as is or if additional changes are required.

core/src/main/java/eu/fasten/core/data/ExtendedRevisionJavaCallGraph.java
amottier: Ok, thanks for the clarification.

core/src/main/java/eu/fasten/core/data/JSONUtils.java
ashkboos: fixed 2b8d727

core/src/main/java/eu/fasten/core/data/JSONUtils.java
ashkboos: fixed b98c359

core/src/main/java/eu/fasten/core/data/JSONUtils.java
ashkboos: fixed afdb9bf

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/data/JSONUtilsTest.java
ashkboos: I would prefer to do some heavy testings by comparing a large number of maven packages JSON strings, for that I selected 100 random maven coordinates. Creating a call graph out of the maven coordinates feature is only available in the opal plugin. If I want to use this feature from the core I would have to declare a dependency to the opal plugin and it would be a cyclic dependency.

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/data/JSONUtilsTest.java
ashkboos: fixed de5c7d9

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/data/JSONUtilsTest.java
amottier: Go it, thanks for the clarification.

amottier:
@ashkboos on my side I think it's all ok for merging. Can I let you do the merge?

ashkboos:
> @ashkboos on my side I think it's all ok for merging. Can I let you do the merge?
@amottier thanks for your review, then I will resolve the conflicts and merge the PR

PR #135, opened 2020-11-25
Update .gitignore with missing Eclipse files

tmortagne:


.gitignore
```diff
@@ -35,6 +35,10 @@ dependency-reduced-pom.xml
 buildNumber.properties
 .mvn/timing.properties
 .mvn/wrapper/maven-wrapper.jar
+.classpath
+.project
+.settings
```
amottier, line 40: Just to be fussy as `.settings` is a folder it should be `.settings/`

amottier (approved):
Looks good to me

PR #179, opened 2020-12-18
Using JSONUtils to send/save very large call graphs

mir-am:
## Description
The newly-developed`JSONUtils` in `eu.fasten.core.data` is used to send and save very large call graphs produced by the OPAL plugin. Specifically, I have made the necessary changes to `CallGraphUtils` and the OPAL plug-in itself.

## Motivation and context
Previously, we encountered quite a number of `OutOfMemoryError`exceptions when converting very large ERCGs objects to a JSON string.

## Testing
- The `Main` module of the OPAL plug-in works fine while saving a very large CG, i.e. `org.jetbrains.kotlin:kotlin-compiler:1.4.21`
- The FASTEN server can now send or save very large CGs on the disk.

## Task list  
- [x] Make changes to `CallGraphUtils.write()` and its usages to make it work with `JSONUtils.toJSONString()`.
- [x] Use `JSONUtils.toJSONString()` in the `Main` module and the `OPAL` pluging class.


gousiosg:
What is the status of this one?

ashkboos (approved):

PR #183, opened 2021-01-04
Fix Kafka configuration for OPAL (rebalancing issue)

wzorgdrager:
## The problem
We noticed that every now and then (read: almost every hour) OPAL consumption would go idle due to [Kafka rebalancing](https://medium.com/streamthoughts/apache-kafka-rebalance-protocol-or-the-magic-behind-your-streams-applications-e94baf68e4f2) its consumer group. Sometimes rebalancing took an hour, causing heavy disruption in the consuming process.
cg-generation

Generally, there are 2 situations when a rebalance is triggered:

A consumer _joins_ a group (and therefore causes a rebalance):
- when an application is (re)started; this happens all the time for OPAL and is inevitable since some CG generations will cause OoM.
- when an application is scaled up; only manually OPAL pods will be scaled upwards.

A consumer _leaves_ a group (and therefore causes a rebalance):
- `max.poll.interval.ms` is exceeded, which means that the polled records are not processed in time. CG generation can take a long time, so this value should be as high as the maximum time spent on generating a CG.
- `session.timeout.ms` is exceeded, which means that the heartbeat thread is not sending heartbeats anymore.
- consumer shuts down (pod relocation, app scale-down, etc.)

Another reason why Kafka rebalances a consumer group could be a ZK session timeout (see [here](https://stackoverflow.com/questions/30988002/what-does-rebalancing-mean-in-apache-kafka-context)). We won't consider that scenario for this PR, but should be taken into account if issues still remain.

When a group is rebalancing, partitions get reassigned and current partitions get revoked. However, a partition gets revoked _only_ when that consumer is done processing its current poll. Since OPAL has (potentially) high processing times it can take up to `max.poll.interval.ms` before OPAL is done and a partition gets revoked. Currently `max.poll.interval.ms` is set to `1 hour`, so this is how long a rebalance might take. We can't change the rebalance protocol nor the time OPAL takes to process a poll, so we want to minimize the amount of rebalances.

## Proposed solution
Kafka Static Membership with a Kubernetes Stateful set. Kafka Static Membership allows for (static) consumers whose identities are persisted and recognized when restarted. This means that when a consumer restarts it gets assigned the _same_ partitions and no rebalancing is happening unless the `session.timeout.ms` is exceeded. We already use static membership in the current setup, but without a K8s stateful set. We need static pod id's and a persisted state, which is not guaranteed for a Deployment (which we currently use). More explicitly, I think we should use the following configuration/setup:
- Kafka Static Membership (this is already implemented)
- Deployment as a Stateful set, see [the PR in the deployment repo](https://github.com/fasten-project/k8s-deployments/pull/9).
- `max.poll.interval.ms` still set to `1 hour`. This is inevitable since we can't influence OPAL's speed.
- `session.timeout.ms` should be `5 minutes`. This is the time after which a consumer is considered 'dead' (i.e. it did not receive heartbeats in that period) and a rebalance is triggered. Moreover, this is the time a pod can safely restart/crash without triggering a rebalance.
- `heartbeat.interval.ms` to `10 seconds`. This is the interval at which heartbeats are sent to the coordinator, but also used to inform a rebalance is in progress. If this value is too high, healthy consumers are taking longer to be aware of a rebalance (currently it's set to 3 minutes). 

## Some considerations
- Default Kafka (consumer) configurations are applied in the Kafka server, which means that every plugin/analyzer will get the same configuration. Since OPAL's config is rather specific, we might want to override the default configuration for OPAL and create a more generic default configuration (without static membership).
- When we deploy a set of OPAL pods, not every application is started at exactly the same time. Therefore many rebalances could be triggered because consumers keep on joining the group (even with static membership). Therefore Kafka has the `group.initial.rebalance.delay.ms` configuration, which decides the initial rebalance delay. This is currently set to `0 ms`. I updated it to `10 seconds` which seems more reasonable. Though, it requires a full cluster restart which might not be preferable at the moment. 

## Task list
- [x] Change `Deployment` to `Stateful set`. See [this PR](https://github.com/fasten-project/k8s-deployments/pull/9).
- [x] Default Kafka config
    - Default heartbeat interval: `5 seconds`
    - Default session timeout: `1 minute`
    - Default max poll interval: `2 minutes`
- [x] Custom plugin configs
  - [x] Custom OPAL config implementing the above mentioned changes.
  - [x] Custom Metadata config: max poll interval of `15 minutes`.
  - [x] Custom Quality Analyzer config:  max poll interval of `1 hour`.
  - [x] Custom Vulnerability Consumer config: max poll interval of `1 hour`.
  - [x] Custom RepoCloner config: max poll interval of `30 minutes`.
  - [x] All other plugins fall back to all default consumer configs. 
- [x] Being able to override consumer group by using `-cg` or `--consumer_group`
- [x] Being able to override output topic by using `-ot` or `--output_topic`. Note: this value is still subsituted in `fasten.OUTPUT_TOPIC.[out|err]`
- [ ] Test?

## Testing
I still have to figure out how to properly test this.

## References
I based my ideas and observations on reading the [Kafka configuration reference](https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html), [this blog post covering a similar issue](https://medium.com/bakdata/solving-my-weird-kafka-rebalancing-problems-c05e99535435g), [a blog post on Kafka rebalancing](https://medium.com/streamthoughts/apache-kafka-rebalance-protocol-or-the-magic-behind-your-streams-applications-e94baf68e4f2),  [a stackoverflow post on rebalancing](https://stackoverflow.com/questions/30988002/what-does-rebalancing-mean-in-apache-kafka-context) and some internal discussions over the past months.

### DIFF
--- a/analyzer/graph-plugin/src/main/java/eu/fasten/analyzer/graphplugin/GraphDatabasePlugin.java
+++ b/analyzer/graph-plugin/src/main/java/eu/fasten/analyzer/graphplugin/GraphDatabasePlugin.java
@@ -28,10 +28,9 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
+
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.json.JSONTokener;
@@ -194,6 +193,11 @@ public class GraphDatabasePlugin extends Plugin {
         public void freeResource() {
             rocksDao.close();
             rocksDao = null;
+        }
+
+        @Override
+        public Properties getConsumerProperties() {
+            return new Properties(); // GraphDB plugin relies on default configuration.
         }
     }
 }
--- a/analyzer/graph-plugin/src/main/java/eu/fasten/analyzer/graphplugin/GraphDatabasePlugin.java
+++ b/analyzer/graph-plugin/src/main/java/eu/fasten/analyzer/graphplugin/GraphDatabasePlugin.java
@@ -194,10 +194,5 @@ public class GraphDatabasePlugin extends Plugin {
             rocksDao.close();
             rocksDao = null;
         }
-
-        @Override
-        public Properties getConsumerProperties() {
-            return new Properties(); // GraphDB plugin relies on default configuration.
-        }
     }
 }
--- a/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/OPALPlugin.java
+++ b/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/OPALPlugin.java
@@ -27,10 +27,9 @@ import eu.fasten.core.data.Constants;
 import eu.fasten.core.data.ExtendedRevisionJavaCallGraph;
 import eu.fasten.core.plugins.KafkaPlugin;
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
+
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.json.JSONObject;
 import org.pf4j.Extension;
 import org.pf4j.Plugin;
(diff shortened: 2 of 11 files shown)


gousiosg:
My only concern is that you make the analysers depend on a library they do not need. Perhaps it may make sense to replace the generic `getConsumerConfig` method to a more specific `getMaxConsumeTimeout` that returns just the value you need.

wzorgdrager:
@gousiosg I understand, but by changing it to `getMaxConsumeTimeout` you 1) can't change/override any other value in the consumer config (this would already be a problem for OPAL because we override multiple consumer configs) and 2) you force analyzers to pick a value for `MaxConsumeTimeout` whereas for `getConsumerConfig` you can just return an empty `Properties` instance and fall back on the default config. Preferably, we want to analyzers have the _possibility_ to override `getConsumerConfig` but since we use an interface rather than an abstract class we can't have this behavior.

gousiosg:
1. You can have multiple such methods (I think you need max 3). In any case, you don't want the analyser to set arbitrary config keys in the returned property object, so you need to check the returned object before you merge it. You might as well have the compiler do the check for you.
2. You can use default implementations for the interface method: https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html

wzorgdrager:
Ah, I wasn't aware of default implementations for interface methods. In that case, I agree! I will update it so only some specific configurations can be overriden. Thanks :)

wzorgdrager:
Will merge for now and do a test run in the coming days.

PR #190, opened 2021-01-11
WIP Consume timeout for FastenKafkaPlugins

wzorgdrager:
## Description
I added a timeout for the consumer part of a Kafka plugin. This will run the consumer part in a different thread and kill it if it did not process it in time. For example, the OPAL plugin might take hours to process on CG (and probably will run OoM eventually) adding a timeout of x minutes will prevent that.

It can be configured by adding this flag to the startup arguments:
`--consume_timeout [TIME_IN_SECONDS]`

If the plugin is working on one record for too long and the timeout is reached, the [thread will be interrupted](https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html) as done on [this line](https://github.com/fasten-project/fasten/blob/3795681682f3f8d25b3ac15e5eab20dddd253d64/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java#L428). However, I'm not sure if this will actually stop the processing. For OPAL I think it does, since [this line](https://github.com/fasten-project/fasten/blob/b07b6450ef24561c1add313eac431d2d8435d515/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/CallGraphConstructor.java#L74) catches all exceptions including the `InterruptedException`. I'm not sure if this is the case for other plugins nor do I know how to handle this properly and ensure the consume thread (i.e. the thread that calls `consume(r.value())` for a particular record) is killed so that I can move onto the next record. 

## Motivation and context
Plugins are working on 1 record for hours which is not desirable. It causes Kafka problems in the form of rebalances. 

## Testing
I did some tests using Mockito. However, it is hard to predict what happens for certain plugins.

## Task list  
- [ ] Test on the cluster with OPAL

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -27,6 +27,9 @@ The FASTEN-server is a component necessary for running [FASTEN-specific plugins]
     - `-pl` `--plugin_list` List of plugins to run. Can be used multiple times.
     - `-po` `--plugin_output` Path to directory where plugin output messages will be stored
     - `-pol` `--plugin_output_link` HTTP link to the root directory where output messages will be stored
+- `cg` `--consumer_group` Name of the consumer group. Defaults to (canonical) name of the plugin.
+- `ot` `--output_topic` Name of the output topic. Defaults to (simple) name of the plugin.
+- `ct` `--consume_timeout` Adds a timeout on the time a plugin can spend on its consumed records. Disabled by default.
 - `-V` `--version` Print version information and exit.
 
 ## Usage 
--- a/server/src/main/java/eu/fasten/server/FastenServer.java
+++ b/server/src/main/java/eu/fasten/server/FastenServer.java
@@ -131,6 +131,13 @@ public class FastenServer implements Runnable {
     )
     String outputTopic;
 
+    @Option(names = {"-ct", "--consume_timeout"},
+            paramLabel = "consumeTimeout",
+            description = "Adds a timeout on the time a plugin can spend on its consumed records. Disabled by default",
+            defaultValue = "-1"
+    )
+    long consumeTimeout;
+
     private static final Logger logger = LoggerFactory.getLogger(FastenServer.class);
 
     @Override
@@ -236,7 +243,9 @@ public class FastenServer implements Runnable {
             return new FastenKafkaPlugin(consumerProperties, producerProperties, k, skipOffsets,
                     (outputDirs != null) ? outputDirs.get(k.getClass().getSimpleName()) : null,
                     (outputLinks != null) ? outputLinks.get(k.getClass().getSimpleName()) : null,
-                    (outputTopic != null) ? outputTopic : k.getClass().getSimpleName());
+                    (outputTopic != null) ? outputTopic : k.getClass().getSimpleName(),
+                    (consumeTimeout != -1) ? true : false,
+                    consumeTimeout);
         }).collect(Collectors.toList());
     }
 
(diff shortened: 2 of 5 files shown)


server/README.md
```diff
@@ -27,6 +27,9 @@ The FASTEN-server is a component necessary for running [FASTEN-specific plugins]
     - `-pl` `--plugin_list` List of plugins to run. Can be used multiple times.
     - `-po` `--plugin_output` Path to directory where plugin output messages will be stored
     - `-pol` `--plugin_output_link` HTTP link to the root directory where output messages will be stored
+- `cg` `--consumer_group` Name of the consumer group. Defaults to (canonical) name of the plugin.
```
mir-am, line 30: Why `--consumer_group` argument for the server?

server/README.md
```diff
@@ -27,6 +27,9 @@ The FASTEN-server is a component necessary for running [FASTEN-specific plugins]
...
+- `ot` `--output_topic` Name of the output topic. Defaults to (simple) name of the plugin.
```
mir-am, line 31: The arg. `-kt` already accepts the output topic for plug-ins. So what's the purpose of `ot`?

server/src/main/java/eu/fasten/server/FastenServer.java
```diff
@@ -236,7 +243,9 @@ private void waitForInterruption(List<FastenServerPlugin> plugins) {
             return new FastenKafkaPlugin(consumerProperties, producerProperties, k, skipOffsets,
```
mir-am, line 243: It seems that `enableKafka` flag is not passed here.

wzorgdrager: I added another constructor which defaults `enableKafka` to false. I indeed need this for test code. I wanted to be able to isolate parts of the FastenKafkaPlugin to test, without starting Kafka consumer/producer. Since these are initialized in the constructor, I couldn't find an easier workaround.

server/README.md
wzorgdrager: Because it was not configurable before and always set to the name of the class. Therefore I couldn't deploy OPAL again for a test run without using a different consumer groud. It would use the same consumer group as used for production + the same offsets (which I want to prevent for a test run). Btw this change was already implemented in [this PR](https://github.com/fasten-project/fasten/pull/183).

server/README.md
wzorgdrager: This change was also already implemented in [this PR](https://github.com/fasten-project/fasten/pull/183). `-kt` is not sufficient, because it only sets the consume topic for a plugin (see [here](https://github.com/fasten-project/fasten/blob/a3dc8f3fd29d493df0782bfa132971a1d23d71db/server/src/main/java/eu/fasten/server/FastenServer.java#L235)). Not the output/produce topic. The output topic was always defaulting to  `k.getClass().getSimpleName()`.

server/src/main/java/eu/fasten/server/FastenServer.java
mir-am: Shouldn't the default value be `true`? if the flag is not passed in `FastenServer.java`?

wzorgdrager: Yes this is the case, see [here](https://github.com/fasten-project/fasten/blob/abe026545bab72ed54290047525279dbcea70471/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java#L111)

server/README.md
mir-am: I see! Perhaps, we need to pass a list like `-kt` for `-ot`? Now, you can only set one output topic assuming that you want to run only one plug-in, which is the case for current deployment.

wzorgdrager: Yes, maybe.. however, what is the chance that FastenServer will ever run more than 1 plugin? I'm not sure if we still want that. I do agree, that we might need to pick one standard and use that for all arguments (either accept for multiple plugins or just for one). I think we should do that in a different PR though

PR #199, opened 2021-01-18
at-least-once semantics and progress tracking in local storage

wzorgdrager:
## Description
This PR covers two improvements:
1. Change default behaviour to **at-least-once** semantics (currently it is **at-most-once**). This means that we always first consume, produce and _then_ commit its offset. This way, we will never lose a record due to shutdown, crashes or errors.
2. Consequently, I also implemented a feature described as __LocalStorage__. Each record that is consumed, is first hashed (using [SHA-1](https://en.wikipedia.org/wiki/SHA-1)) and stored in local storage. Once it's properly processed and the output is sent to the producer, the hashed record is removed from this local storage. This prevents a scenario where a pod crashes during processing, restarts, and then start re-processing that same record (which might result in a deadlock for unprocessable records). We need this kind of behaviour to allow **at-least-once** semantics. 
	
     Requirements to use LocalStorage:
		    - Enable LocalStorage using the `--local_storage` flag.
		    - Use `---local_storage_dir` flag to define output directory. Ensure that this directory is accessible by all pods on all servers (e.g. NFS mount). 
		  - Configure environment variable `POD_INSTANCE_ID`. This needs to be static and unique per pod (i.e. use a StatefulSet deployment).
		
The full commit protocol using local storage is as follows:
```
1. Poll one record (by default).
2. If the record hash is in local storage:
     a. Produce to error topic (this record is probably processed before and caused a crash or timeout).
     b. Commit the offset, only if the producer confirmed sending the message.
     c. Delete record in local storage.
     d. Go back to 1.
3. If the record hash is _not_ in local storage:
     a. Process the record.
     b. Produce its results (either to the error topic, or output topic).
     c. Commit the offset, only if the producer confirmed sending the message.
     d. Delete record in local storage.
     e. Go back to 1.
```

__Some notes:__
-  Most of our plugins are not fast; they spend quite some time processing only one record. If you have a plugin which is very fast, local storage might slow this down (reads and writes to NFS mount can be considered expensive). 
- Local storage is only tested with batch size 1 (this is also the default for all our plugins). 
## Motivation and context
Currently, we have **at-most-once** semantics which gives us no guarantees if all records will be processed. Moreover, if we want to support **at-least-once** we have to deal with artifacts that crash a pod (for example for OPAL OoM errors). Local storage allows us to discard/skip records that we processed before and caused an issue. 

## Testing
I tested the LocalStorage feature in a unit test. The new commit strategy is tested using some (simplified) mock scenario. It still requires a test on the cluster. 

## Task list  
- [ ] Test on cluster for OPAL.

### DIFF
--- a/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
+++ b/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
@@ -34,6 +34,7 @@ import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.commons.lang.StringUtils;
 import org.apache.kafka.clients.consumer.CommitFailedException;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
@@ -80,7 +81,7 @@ public class FastenKafkaPlugin implements FastenServerPlugin {
      * @param skipOffsets        skip offset number
      */
     public FastenKafkaPlugin(boolean enableKafka, Properties consumerProperties, Properties producerProperties,
-                             KafkaPlugin plugin, int skipOffsets, String writeDirectory, String writeLink, String outputTopic, boolean consumeTimeoutEnabled, long consumeTimeout, boolean exitOnTimeout) {
+                             KafkaPlugin plugin, int skipOffsets, String writeDirectory, String writeLink, String outputTopic, boolean consumeTimeoutEnabled, long consumeTimeout, boolean exitOnTimeout, boolean enableLocalStorage) {
         this.plugin = plugin;
 
         if (enableKafka) {
@@ -100,8 +101,12 @@ public class FastenKafkaPlugin implements FastenServerPlugin {
                     ? writeLink.substring(0, writeLink.length() - 1) : writeLink;
 
 
-            // TODO Check if local storage needs to be enabled.
-            this.localStorage = new LocalStorage(this.writeLink);
+            // If the write link is not null, and local storage is enabled. Initialize it.
+            if (enableLocalStorage) {
+                this.localStorage = new LocalStorage(this.writeLink);
+            } else {
+                this.localStorage = null;
+            }
         } else {
             this.writeLink = null;
             this.localStorage = null;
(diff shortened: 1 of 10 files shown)


gousiosg:
I don't have any comments on this PR, LGTM.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -161,37 +168,84 @@ public void stop() {
...
-    private void handleConsuming() {
-        ConsumerRecords<String, String> records = connection.poll(Duration.ofSeconds(1));
-        for (var r : records) {
-            doCommitSync();
-            var consumeTimestamp = System.currentTimeMillis() / 1000L;
+    public void processRecord(ConsumerRecord<String, String> record, Long consumeTimestamp) {
+        if (localStorage != null) { // If local storage is enabled.
+            if (localStorage.exists(record.value())) { // This plugin already consumed this record before, we will not process it now.
+                logger.info("Already processed record with hash: " + localStorage.getSHA1(record.value()) + ", skipping it now.");
+                plugin.setPluginError(new ExistsInLocalStorageException("Record already exists in local storage. Most probably it has been processed before and the pod crashed."));
```
mir-am, line 214: Do you think that it might be a good idea to add the input record to the plugin's error in the case of `ExistsInLocalStorageException`? Consider a crash case where neither output nor error is produced, e.g. OOM.

mir-am (commented):
Overall, the PR looks good to me and it's a nice solution. Have you tested this solution for the OPAL deployment yet?

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: > Do you think that it might be a good idea to add the input record to the plugin's error in the case of `ExistsInLocalStorageException`? Consider a crash case where neither output nor error is produced, e.g. OOM.

@wzorgdrager any thoughts or ideas?

wzorgdrager: Yes I agree, but I thought this was already the case when it was outputted to the error topic. I will take a look.

wzorgdrager: The input is already added in this line (in case there is an exception) https://github.com/fasten-project/fasten/blob/fe64da43b641bca3c83848fffdf462c4ea822ab2/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java#L311 I will merge for now, to test on the cluster.

PR #208, opened 2021-01-28
Utility parser for partial FASTEN URI

### LINKED ISSUE #207: Utility to parse FASTEN partial uri
## Is your feature request related to a problem? Please describe.
There is a utility tool that parses Full FASTEN URI into 3 parts: package name, forge, and partial URI. Partial URI by itself is not atomic and can be split into more parts such as namespace, class, method name, and its signature. More specifically, this is needed for the frontend where a user-friendly text is wanted, instead of an unreadable URI.

## Describe the solution you'd like
Implement a separate utility function that would parse partial URI into namespace, class name, method name, and its signature.

romatallinn:
Closes #207

## Description
PR brings a new utility method to `core/src/main/java/eu/fasten/core/utils/FastenUriUtils.java` that parses partial FASTEN uri.

For example,

**INPUT:** `/junit.awtui/AboutDialog.<init>(/java.awt/Frame)/java.lang/VoidType`

**OUTPUT:**
- Namespace: `junit.awtui`
- Class: `AboutDialog`
- Method: `init`
- Method Args: `/java.awt/Frame`
- Method Return Type: `/java.lang/VoidType`

## Task list  
- [x] Implement parser
- [x] Write tests for the parser
- [x] Write tests for other methods in FastenUriUtilis

### DIFF
--- a/core/src/main/java/eu/fasten/core/utils/FastenUriUtils.java
+++ b/core/src/main/java/eu/fasten/core/utils/FastenUriUtils.java
@@ -76,18 +76,19 @@ public class FastenUriUtils {
         }
 
         // URI is usually stored encoded, so decode for parsing.
+        // E.g., `/com.sun.istack.localization/Localizer.%3Cinit%3E(%2Fjava.util%2FLocale)%2Fjava.lang%2FVoidType`
         partialFastenUri = java.net.URLDecoder.decode(partialFastenUri, StandardCharsets.UTF_8);
 
         // Namespace: `/{namespace}/`
         Pattern namespacePattern = Pattern.compile("(?<=/)(.+?)(?=/)");
         Matcher namespaceMatcher = namespacePattern.matcher(partialFastenUri);
-        if (!namespaceMatcher.find())
+        if (!namespaceMatcher.find() || namespaceMatcher.group(0).isEmpty())
             throw new IllegalArgumentException(partialUriFormatException);
 
         // Class: `/{class}.*(`
         Pattern classPattern = Pattern.compile("(?<=/)([^\\/]+)(?=\\.([^./]+)\\()");
         Matcher classMatcher = classPattern.matcher(partialFastenUri);
-        if (!classMatcher.find())
+        if (!classMatcher.find() || classMatcher.group(0).isEmpty())
             throw new IllegalArgumentException(partialUriFormatException);
 
 
--- a/core/src/main/java/eu/fasten/core/utils/FastenUriUtils.java
+++ b/core/src/main/java/eu/fasten/core/utils/FastenUriUtils.java
@@ -1,5 +1,6 @@
 package eu.fasten.core.utils;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


romatallinn:
@MihhailSokolov, I don't have configured access to the production instance. Can you please give me a favour: look up a real example of a callable uri in DB. I just noticed that API is returning URIs with partial encoding (some symbols in uri are encoded, others are not). So, if they are actually stored like this in DB and operated by the system as such, then this utility method will not work. I then need to decode the string first. 

So, I want to see how a callable uri looks straight from the DB. Because it might be that API plugin somehow encodes it when returns.

MihhailSokolov:
> @MihhailSokolov, I don't have a configured access to the production instance of DB. Can you please look up for me a real example of a callable URI straight from DB (not from API). I just noticed that API is returning URIs with partial encoding (some symbols in uri are encoded, others are not). So, if they are actually stored like this in DB and operated by the system as such, then this utility method will not work. I then need to decode the string first. 
> 
> So, I want to see how a callable uri looks straight from the DB. Because it might be that API plugin somehow encodes it when returns.

Here is an example of 5 FASTEN URIs:
```
 /com.sun.istack.localization/Localizer.localize(Localizable)%2Fjava.lang%
2FString                                                                  
 /com.sun.istack.localization/Localizer.getLocale()%2Fjava.util%2FLocale  
 /com.sun.istack.localization/Localizer.getDefaultMessage(Localizable)%2Fj
ava.lang%2FString                                                         
 /com.sun.istack.localization/Localizer.%3Cinit%3E(%2Fjava.util%2FLocale)%
2Fjava.lang%2FVoidType                                                    
 /com.sun.istack.localization/Localizer.%3Cinit%3E()%2Fjava.lang%2FVoidTyp
e
```
If you want to learn more about the FASTEN URI specification, you can read up this document https://www.fasten-project.eu/view/wiki/WP1/FASTEN%20URI%20scheme%20%28Proposal%29

romatallinn:
@ashkboos or @MihhailSokolov could you please have a quick look through the PR?

ashkboos:
I guess we already have the namespace functionality in the `FastenUri` class!

romatallinn:
> I guess we already have the namespace functionality in the `FastenUri` class!

That's true, although partial one does have it as well, so I just parse it as is. Do you want to remove this matcher?

ashkboos:
> > I guess we already have the namespace functionality in the `FastenUri` class!
> 
> That's true, although partial one does have it as well, so I just parse it as is. Do you want to remove this matcher?

It is safer to use the logic that we already had and tested. But then your code will not be consistent if you use that functionality for one part and use the matcher for other parts. So I leave it to you.

romatallinn:
> I guess we already have the namespace functionality in the `FastenUri` class!

Actually, I thought you meant that full FASTEN URI parses namespace as well, but I just understood that it doesn't. I see you meant the actual FastenURI object, and yes indeed it has namespace in it. But at least for the purpose, I need this parser for, it is much easier to parse namespace along with other parameters. So, if it is up to me, then I guess I'll leave it here.

If you don't have any other comments, can you please approve the PR?

ashkboos (commented):
I suggest adding some extra logic to your tests that use `FastenUri.create(String uri)` to make sure if the uris are correct.

PR #214, opened 2021-01-31
Set Kafka max message size to 50MB

wzorgdrager:
Addresses issue as reported here: https://github.com/fasten-project/fasten-docker-deployment/issues/1

Set producer and consumer side max message size to 50MB. This is in line with the broker configuration.


MagielBruntink:
Thanks this fixes the issue.

PR #219, opened 2021-02-01
API: Add module exist check for endpoints

### LINKED ISSUE #213: API: module's callables endpoint returns 200 even if module doesn't exist
## Describe the bug
The endpoint that is meant to return the list of callables for a specific module in the package returns successful response (w/ empty set) even if module doesn't exist for the package.

## To Reproduce
**POST** `https://api.fasten-project.eu/api/mvn/packages/jboss:jbossmq-client/3.2.3/modules/callables`
With body: `abcde` (random string)

-> returns 200

## Expected behavior
The endpoint shall return `404` if module is not found for the package.

romatallinn:
Closes #213 

## Description
PR adds check for module existence in the endpoints of internal elements.  In the current implementation, if the module doesn't exist, such endpoints would still return `200` with an empty set, while supposed `404`.

For example, retrieval of the callables for a module, `/{pkg}/{pkg_ver}/modules/callables`.

### DIFF
--- a/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
+++ b/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
@@ -1500,17 +1500,7 @@ public class MetadataDao {
                                   int offset,
                                   int limit) {
 
-        // SQL query
-        /*
-            SELECT {e.* | (e.source_id, e.target_id)}
-            FROM edges AS e
-                JOIN callables AS c ON e.source_id = c.id
-                JOIN modules AS m ON m.id = c.module_id
-                JOIN package_versions AS pv ON pv.id = m.package_version_id
-                JOIN packages AS p ON p.id = pv.package_id
-            WHERE p.package_name = <packageName>
-                AND pv.version = <packageVersion>;
-         */
+        if(!assertPackageExistence(packageName, packageVersion)) return null;
 
         // Tables
         Packages p = Packages.PACKAGES;
--- a/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
+++ b/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
@@ -907,6 +907,20 @@ public class MetadataDao {
                 .and(PackageVersions.PACKAGE_VERSIONS.VERSION.equalIgnoreCase(version));
     }
 
+    protected boolean assertPackageExistence(String name, String version) {
+        Packages p = Packages.PACKAGES;
+        PackageVersions pv = PackageVersions.PACKAGE_VERSIONS;
+
+        Record selectPackage = context
+                .select(p.fields())
+                .from(p)
+                .innerJoin(pv).on(p.ID.eq(pv.PACKAGE_ID))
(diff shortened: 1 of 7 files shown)


romatallinn:
I found out that there are quite a lot of endpoints that had the same issue. I hope I added the necessary checks for all of them. So if a wrong package or module will be requested by the user, API will not return just an empty set, but actually will throw an error.

But @MihhailSokolov I'd appreciate your opinion on whether or not my solution is optimal. In particular, `assertPackageExistence` and `assertModulesExistence` in `MetadataDao.java` are the methods that check for it. Do you know maybe there is a more elegant way to assert it within the main query? Or this solution will work?

MihhailSokolov:
Sure, I will review it either later today or tomorrow. I would really appreciate it if you could also increase a timeout for all requests while we are at it. This should be as easy as setting some properties in the `application.properties` of the REST API. Thank you!

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
```diff
@@ -1056,6 +1089,9 @@ public String getPackageModules(String packageName,
                                     String packageVersion,
                                     int offset,
                                     int limit) {
+
+        if(!assertPackageExistence(packageName, packageVersion)) return null;
```
MihhailSokolov, line 1093: This way we need to run 2 queries whenever the package or modules are retrieved. This might affect the performance (not in a good way). I think a better solution be to prepend the `count(<retrieved field>)` to select and just check if it is > 0 before returning the result and if <1 return `null`, i.e. in this case the query would be something like `select count(f.*), f.* from ....`.
Same applies to the other changed methods.

MihhailSokolov (approved):
See the comment. I think it is not the best solution to run 2 queries each time we want to retrieve packages or modules. The existence can be checked in the same query using `count(...)` statement. I would also appreciate if you could increase the timeout for the REST API.
Thank you!

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
romatallinn: > a better solution be to prepend the `count(<retrieved field>)`
I surely thought about this way, but it won't work as expected; an empty result can in fact happen.

> to run 2 queries whenever the package
But I completely agree that this sucks. That's why I asked you for advice on whether or not there is an alternative.

MihhailSokolov: > > a better solution be to prepend the `count(<retrieved field>)`
> 
> I surely thought about this way, but it won't work as expected; an empty result can in fact happen.
> 
> > to run 2 queries whenever the package
> 
> But I completely agree that this sucks. That's why I asked you for advice on whether or not there is an alternative.
> 
> Any thoughts?

Hm, yes, I guess you are right. Then will have to go with 2 queries for now as you implemented. Feel free to merge this when you increase the timeout. Thanks!

romatallinn: > Feel free to merge this when you increase the timeout.

Moved to #277

PR #228, opened 2021-02-04
Reducing the size of the Docker image for FASTEN dev.

mir-am:
## Description
In this PR, I try to reduce the size of the FASTEN's dev. image and the JAR files of plugins as much as possible. Also, we need to make sure that we will not encounter any issues as a result of reducing the size of images and removing unnecessary dependencies.

Specifically, I made changes to the following files:
- `fasten/docker/server/Dockerfile`: Added `openjdk:11-jre-slim-buster` for reducing the size of the image.
- `fasten/analyzer/javacg-opal/pom.xml`: Enabled `<minimizeJar>true</minimizeJar>` for reducing the size of the JAR file.

## Motivation and context
As pointed out by the partners, the size of the FASTEN dev. image has recently increased, i.e., 1.2GB. By using the slim version of the OpenJDK and minimizing JARs, it is possible to reduce the size of the image significantly.

## Testing
- The FASTEN server image should be deployed and tested to ensure that it will not have issues after reducing the size of its image.
- The OPAL plugin should be tested using its `Main` class and the FASTEN server.

## Task list  
- [x] Add `openjdk:11-jre-slim-buster` to the FASTEN server's Docker file
- [x] Enable minimize JAR option for the OPAL plugin in its POM file
- [x] Test the FASTEN server with various plugins with the new image
- [x] Test the OPAL plugin after minimizing its JAR file

## Additional context
- The size of the FASTEN dev. image is reduced by around 60%. Its size is now around **730MB**.
- The size of the OPAL plugin's JAR file is reduced by 50%. Its size is now 99MB.

### DIFF
--- a/.github/workflows/dockerpublish_develop_slim.yaml
+++ b/.github/workflows/dockerpublish_develop_slim.yaml
@@ -1,0 +1,63 @@
+name: Publish Docker image
+
+on:
+  push:
+    # Publish `develop` as Docker `latest` image.
+    branches:
+      - develop
+
+  # Run tests for any PRs.
+  pull_request:
+    branches:
+      - develop
+
+env:
+  # TODO: Change variable to your image's name.
+  IMAGE_NAME: fasten.server.develop.slim
+
+jobs:
+
+  # Push image to GitHub Packages.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
+
+    runs-on: ubuntu-18.04
+    if: github.event_name == 'push'
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+      - name: Setup JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.11
+      - name: Build jar
+        run: mvn clean install
+      - name: Build image
+        run: cd docker/ && docker build . -f slim/Dockerfile -t fasten.server.develop.slim
+      - name: Log into registry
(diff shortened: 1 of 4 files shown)


MagielBruntink:
I built the image locally for testing: quite a bit smaller, nice work!
```
REPOSITORY                              TAG                  IMAGE ID            CREATED             SIZE
fasten.server.develop-image-dev-size    latest               bc6d4484dad5        8 seconds ago       743MB
```

Then I ran it through docker-compose deployment without any issue, steps below:

1. Checkout your branch image-dev-size in fasten, `mvn clean install -DskipTests`
2. Build the image locally as follows: `$ docker build -t fasten.server.develop-image-dev-size -f docker/server/Dockerfile docker/`
3. Clone https://github.com/fasten-project/fasten-docker-deployment
4. Edit .env to read `FASTEN_SERVER_IMAGE=fasten.server.develop-image-dev-size:latest`
5. `$ source start_mvn_services.sh`
6. `$ cat test-resources/mvn.coords.txt | docker-compose exec -T kafka kafka-console-producer.sh --broker-list kafka:19092 --topic fasten.mvn.test --property 'parse.key=true' --property 'key.separator=|'`
7. Observe Kafka topics at localhost:9000 or http://localhost:9001/mvn/packages/log4j:log4j to see results.

MagielBruntink:
Very nice, thanks!

PR #229, opened 2021-02-04
Repo analyzer

ilyagrishkov:
## Description
Analyser that determines quality of tests in a repository cloned by the RepoCloner plugin. The plugin consumes a record containing a path to the cloned repository and outputs the following stats for every module of the project:
- Number of source files
- Number of test files
- Ratio of source to test files
- Number of functions in source files
- Number of unit tests  
- Ratio of unit tests to functions
- Number of test files that import any mocking framework
- Number of unit tests that use mocking
- Ratio of tests with and without mocking

## Motivation and context
Dynamic call graph generation requires high enough coverage of the project under analysis. This tool aims at identifying metrics that would help to filter out project with low coverage.

## Testing
Unit tests for the analyser, followed by bug fixes after deployment (if any).

## Task list  
- [x] Add support for build managers other than Maven (e.g. Gradle)
- [x] Retrieve number of source and test files according to paths provided in pom.xml or default ones
- [x] Retrieve number of tests for most popular testing frameworks 
- [x] Determine the number of files and unit tests that contain mocking
- [x] Parse source files and get number of functions


ashkboos (commented):
In the plugin you call the analyze method, and in analyze method you only calculate test file number right?

ilyagrishkov:
> In the plugin you call the analyze method, and in analyze method you only calculate test file number right?

In the plugin I create an instance of the analyzer (Maven, Gradle, Ant) and then call the analyze() method that calculates all statistics for each module.

ashkboos:
Did you also check the memory consumption of the plugin? I guess you keep all code in String and analyze them all together? in cases that the project is very big, can it cause memory issues?

ashkboos:
In the method of() in the RepoAnalyzer class, you create instances of different subclasses of RepoAnalyzer. I think this is not a good design practice. Parents shouldn't know about their children (from a design perspective). I think a better solution would be using the Factory design pattern.

ilyagrishkov:
> Did you also check the memory consumption of the plugin? I guess you keep all code in String and analyze them all together? in cases that the project is very big, can it cause memory issues?

It shouldn't cause a problem. It shouldn't take more than a couple (2-3GB) of RAM worst case. Haven't profiled it though, but java files (which are essentially text) are quite lightweight, even if the project is big

ilyagrishkov:
> In the method of() in the RepoAnalyzer class, you create instances of different subclasses of RepoAnalyzer. I think this is not a good design practice. Parents shouldn't know about their children (from a design perspective). I think a better solution would be using the Factory design pattern.

Fixed in f92bc86

ashkboos:
> Fixed in [f92bc86](https://github.com/fasten-project/fasten/commit/f92bc8632eedb419a077533933c02095b4b2af24)
I guess BuildManager can also go to Factory

ilyagrishkov:
> > Fixed in [f92bc86](https://github.com/fasten-project/fasten/commit/f92bc8632eedb419a077533933c02095b4b2af24)
> 
> I guess BuildManager can also go to Factory

Moved it to a separate file

ashkboos:
In general very good code. Thanks a lot. It's ready for merge IMO.

Issue #242, opened 2021-02-12
Convert hard-coded properties in the REST API to K8s ConfigMaps


mir-am:
It is not a good practice to hard-code addresses and properties in the code. In the deployment, we can create K8s ConfigMaps to define env. variables that can be used in the REST API to retrieve required properties such as DB address. This is already done for the metadata plugin.

PR #243, opened 2021-02-12
Add placeholders in Spring app properties instead of hard-coded values

mir-am:
## Description
Spring application properties in `application.properties` file now have placeholders so that properties and addresses can be passed to the REST API as arguments.
This PR addresses Issue #242.

## Motivation and context
It is a good practice to avoid hard-coded values in the code and pass values as arguments to the application.

## Testing
The REST API image should be tested by passing the required arguments.

## Task list  
- [x] Create required K8s configmaps for the REST API properties
- [x] Revise the deployment manifest file of the REST API for including configmaps and passing arguments
- [x] Define placeholders in  `application.properties`

## Additional context
The passed arguments to the Spring application can be checked in the K8s manifest file [here](https://github.com/fasten-project/k8s-deployments/blob/master/fasten/rest-api/restapi-deployment.yaml).


MihhailSokolov (approved):
LGTM, thanks Amir!

PR #251, opened 2021-03-02
Optional drill-down + Extra Info Vulnerability

elanzini:
The drill-down is currently not included in the consumer pipeline, more indexes need to be placed on the DB to accommodate this option. The following fields have been included in the Vulnerability definition: `vectorCVSS2`, `vectorCVSS3`, `cweIds`. Soon the statements stored in the DB will include this information.

### DIFF
--- a/analyzer/vulnerability-consumer/src/main/java/eu/fasten/analyzer/vulnerabilityconsumer/utils/Vulnerability.java
+++ b/analyzer/vulnerability-consumer/src/main/java/eu/fasten/analyzer/vulnerabilityconsumer/utils/Vulnerability.java
@@ -93,6 +93,9 @@ public class Vulnerability {
     private List<String> firstPatchedPurls;
     private Double scoreCVSS2;
     private Double scoreCVSS3;
+    private String vectorCVSS2;
+    private String vectorCVSS3;
+    private List<String> cweIds;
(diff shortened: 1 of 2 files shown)


analyzer/vulnerability-consumer/src/main/java/eu/fasten/analyzer/vulnerabilityconsumer/utils/Vulnerability.java
```diff
@@ -141,6 +144,18 @@ public Double getScoreCVSS3() {
         return scoreCVSS3;
     }
 
+    public String getVectorCVSS2() {
```
mir-am, line 147: Does this new extra information get injected into the metadata field of packages?

mir-am (commented):
Overall, the PR looks good to me. It can be merged.

analyzer/vulnerability-consumer/src/main/java/eu/fasten/analyzer/vulnerabilityconsumer/utils/Vulnerability.java
elanzini: Yes, it will be contained in each Vulnerability Statement that we store.

PR #255, opened 2021-03-18
Merger efficiency improvements

MagielBruntink:
## Description
10x speed improvements and bug fixes for the localMerger.

## Motivation and context
Local merger suffered from call site stitching change, and earlier inefficiencies in data structures + processing.

## Testing
Unit tests that passed before still pass. Added some tests to keep track of runtime on callgraphs of example project (jpacman-framework).


ashkboos:
@MagielBruntink I already looked into all changes during the week, but I took a quick look again now and everything looks good to me.

Issue #256, opened 2021-03-18
Kafka max request size is still too small


MagielBruntink:
The MetadataDBExtension (and maybe other) plugins need a larger max.request.size setting, and then I think it will successfully process fastutil. I suggest we set it to 200 MB.

```
[2021-03-18 20:40:45,644] [INFO ] [main] [e.f.a.m.MetadataDBExtension] - Saved the 'it.unimi.dsi:fastutil:8.5.2' callgraph metadata to the database with package version ID = 912

org.apache.kafka.common.errors.RecordTooLargeException: The message is 75017198 bytes when serialized which is larger than the maximum request size you have configured with the max.request.size configuration.
```

wzorgdrager:
@MagielBruntink The Kafka limit is now set to 50MB and updating this requires all brokers to be restarted. I will resolve this at the beginning of next week!

proksch:
Would it be possible to use the same "trick" that we use for the call-graph generation and serialize the expensive part of the payload to the file system?

MagielBruntink:
I think this message appears because the call graph is indeed send as a Kafka message from the Metadata DB Extension to the Graph DB Extension. It would indeed be better to use the file system JSON of the call graph, if possible.

PR #258, opened 2021-03-23
Increased Kafka max request size to 200 MB.

MagielBruntink:
This should allow big call graphs to be sent from metadata db extension to graph db extension (eg. fastutil).

### DIFF
--- a/server/src/main/java/eu/fasten/server/connectors/KafkaConnector.java
+++ b/server/src/main/java/eu/fasten/server/connectors/KafkaConnector.java
@@ -56,7 +56,7 @@ public class KafkaConnector {
         properties.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
         properties.setProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
         properties.setProperty(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, "1");
-        properties.setProperty(ConsumerConfig.FETCH_MAX_BYTES_CONFIG, "50000000"); //Set max read size to 50 MB.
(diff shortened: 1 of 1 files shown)


wzorgdrager:
This resolves the size issue on the client-side. If we want to resolve this completely, we need to set the same values in all the brokers (which requires a broker restart). As @proksch mentioned in #256, we can also write the payloads to disk. In that case, this PR is unnecessary. I will discuss internally if we can simply restart the metadata plugin with the correct flags to write to disk.

MagielBruntink:
@wzorgdrager @proksch Did you reach a conclusion on this?

proksch:
Sorry, this "internal discussion" has never been initiated. My personal preference would be to follow the example of the call graph generator and serialize the payload to disk.

wzorgdrager:
I actually tried to redeploy it but there is a bug in the code (in short; it seems that there is no proper output path for the metadata plugin which can only be set in the code). I will try to resolve this bug the coming weekend, redeploy and let you know.

wzorgdrager:
An update for now; I've refactored some of the code and redeployed the plugin, however for some reason it does not write to disk yet. It's odd because it does work for OPAL. We have to look into this.

MagielBruntink:
@wzorgdrager There appears to be no code that actually writes the call graph to disk at all, just produce a json gidgraph to its output topic and done.

proksch:
This change was never merged, is it still relevant? After the discussion that we had some days ago in the dev call, it seems that we decided to avoid large messages in the future (esp. the call graphs), so I would assume that we can reject and close this change?

MagielBruntink:
You are right, this should be closed without a merge.

amottier:
@MagielBruntink is it ok to delete [increase-max-kafka-request-size-to-200mb](https://github.com/fasten-project/fasten/tree/increase-max-kafka-request-size-to-200mb) branch?

MagielBruntink:
Done

PR #261, opened 2021-03-26
Include first_patched_purls and Extended lookup for callables

elanzini:
- included the `first_patched_purl` in the published statement
- extended callable lookup to versions before `last_vulnerable` in case nothing is found
- catching exception on parsing of PURLs since some are problematic from the producer. The `getObjectFromPurl` will be refactored soon


mir-am (commented):
Overall, LGTM!
However, we need to discuss a bit about your temp. workaround for callables.

elanzini:
> However, we need to discuss a bit about your temp. workaround for callables.

@mir-am I replaced the workaround with a call to the API to use the `lazy ingest`.

mir-am (approved):
LGTM!

PR #263, opened 2021-03-30
License compliance verifier

### LINKED ISSUE #223: Come up with license compatibility rules
D4.4 expects the [license and compliance plugin](https://github.com/fasten-project/fasten/tree/endocode/compliancePlugin/analyzer/compliance-analyzer) to express compatibility issues when using multiple licenses.
We expect we will need a matrix that pairs licenses and contains compatibility metrics, e.g., an "incompatibility index", a severity level, etc.

The license compatibility matrix would look something like this:

|  | Apache 2 | MIT | GPL 3 | ... |
| -- | -- | -- | -- | -- |
| Apache 2 | &ndash; |   |  |  |
| MIT |   | &ndash; |  |  |
| GPL3 |   |   | &ndash; |  |
| ... |   |   |  |  |

Such a matrix might not be enough to describe more complex compatibility issues among multiple licenses, e.g., a set of licenses excluding another set.
We might need a "license compatibility graph" that stores the "direction" of [in]compatibilities.

If that ends up to be true, the result itself could consist of such a graph, depicting compatibility issues in a graphical form.

marcomicera:
TODO

Closes #223.

### DIFF
--- a/analyzer/compliance-verifier/README.md
+++ b/analyzer/compliance-verifier/README.md
@@ -1,0 +1,35 @@
+<p align="center">
+    <!-- TODO Generate image -->
+</p>
+<br/>
+<p align="center">
+    <a href="https://github.com/fasten-project/fasten/actions">
+        <img src="https://img.shields.io/github/workflow/status/fasten-project/fasten/Java%20CI?logo=GitHub%20Actions&logoColor=white&style=for-the-badge" alt="GitHub Workflow Status"/></a>
+    <!-- Here should be a link to Maven repo and version should be pulled from there. -->
+    <a href="https://github.com/fasten-project/fasten/">
+                <img src="https://img.shields.io/maven-central/v/fasten/opal?label=version&logo=Apache%20Maven&style=for-the-badge" alt="GitHub Workflow Status" /></a>
+</p>
+<br/>
+
+<!-- TODO Improve this -->
+This will only compare licenses and their compliance.
+This plugin will not detect any licenses (done by [#48](https://github.com/fasten-project/fasten/pull/48)).
+
+## Arguments
+<!-- TODO -->
+
+## Usage 
+
+<!-- TODO -->
+
+## Join the community
+
+The FASTEN software package management efficiency relies on an open community contributing to open technologies. Related research projects, R&D engineers, early users and open source contributors are welcome to join the [FASTEN community](https://www.fasten-project.eu/view/Main/Community), to try the tools, to participate in physical and remote worshops and to share our efforts using the project [community page](https://www.fasten-project.eu/view/Main/Community) and the social media buttons below.  
+<p>
+    <a href="http://www.twitter.com/FastenProject">
+        <img src="https://img.shields.io/badge/%20-Twitter-%231DA1F2?logo=Twitter&style=for-the-badge&logoColor=white"  alt="Fasten Twitter" /></a>
+    <a href="http://www.slideshare.net/FastenProject">
+                <img src="https://img.shields.io/badge/%20-SlideShare-%230077B5?logo=slideshare&style=for-the-badge&logoColor=white" alt="GitHub Workflow Status" /></a>
+    <a href="http://www.linkedin.com/groups?gid=12172959">
+            <img src="https://img.shields.io/badge/%20-LinkedIn-%232867B2?logo=linkedin&style=for-the-badge&logoColor=white" alt="Gitter" /></a>
+</p>


lisa-noeth:
Clarifying next steps for the integration here: https://github.com/fasten-project/LCV-CM/issues/1

michelescarlato:
We are working on the integration of LCV within the architecture
[Issues](https://github.com/fasten-project/LCV-CM/issues/1)

lisa-noeth:
We need to define when and where we should run the LCV plugin.

michelescarlato:
Dear @mir-am Amir,

I am trying to understand where and when the License Compliance Verification should happen.

I thought that the best approach to retrieve the list of licenses is to analyze the stitched callgraph. 

Simply because I would need to run the compliance verification right after all the licenses detector scanned the packages.

I can observe which package versions are used from the stitched and hopefully have license information at the file level.

I understand that there is no dedicated plugin for the stitching that can produce an output that could trigger the license compliance verification. Is that correct?

Reading the [End to End REST and Stitching Example](https://github.com/fasten-project/fasten/wiki/End-to-End-REST-and-Stitching-Example#2--stitching), and looking at the stitching section, I am not understanding very well where the stitching action will occur. At server level?

A further question is: will the stitched contains also licenses information?

Thanks,
Michele

mir-am:
> Dear @mir-am Amir,
> 
> I am trying to understand where and when the License Compliance Verification should happen.
> 
> I thought that the best approach to retrieve the list of licenses is to analyze the stitched callgraph.
> 
> Simply because I would need to run the compliance verification right after all the licenses detector scanned the packages.
> 
> I can observe which package versions are used from the stitched and hopefully have license information at the file level.
> 
> I understand that there is no dedicated plugin for the stitching that can produce an output that could trigger the license compliance verification. Is that correct?
> 
> Reading the [End to End REST and Stitching Example](https://github.com/fasten-project/fasten/wiki/End-to-End-REST-and-Stitching-Example#2--stitching), and looking at the stitching section, I am not understanding very well where the stitching action will occur. At server level?
> 
> A further question is: will the stitched contains also licenses information?
> 
> Thanks,
> Michele

Hi Michele,
I would like to refer you to our colleague Mihhail (@MihhailSokolov) as he knows better about stitched CGS and metadata DB. 
He may be able to help you out with your questions and the integration of LCV.

MihhailSokolov:
@michelescarlato
Since license verification requires all the package versions in the dependency set to have been processed with license detector, the verifier cannot go into any point in the pipeline because the whole pipeline is designed to process only one package version at a time. The verification should happen on user request (for example, through the REST API). At this point I would just create a class which you can instantiate, call a method that would check license compliance for certain artifact (and its dependency set) and return the result. Then this result can be used in the REST API or wherever else needed.
As you rightfully noticed, you will need to stitch the dependency set. The result of the stitching is a `DirectedGraph` which consists of only IDs (nodes) and pairs of IDs (edges) and does not carry any metadata. From this directed graph, you will get IDs of all nodes, and you can use them to retrieve the metadata from metadata database since IDs in the graph are the same as `callables.id` in the database.
I hope I've cleared things up a little bit. Do let me know if you have any more questions.

michelescarlato:
Thanks, both @mir-am, and @MihhailSokolov!

As I observed the stitch call graph generated by pycg-stitch, I probably lacked information regarding the stitched one generated in Java.

Could you please point me to some examples?

As far as I could observe from the [output of pycg-stitch](https://github.com/fasten-project/LCV-CM/blob/develop/src/LCV/json/benchmark-pycg-stitch/macro-out.json#L21148), in python, there is a part of the stitched callgraph that contains package versions and related metadata.
@vitsalis, am I correct with this interpretation?


Thanks in advance to all of you!

MihhailSokolov:
> Could you please point me to some examples?

Here is an example that is quite similar to yours:
https://github.com/fasten-project/fasten/blob/ce906724d2bfb85c8f6a943835ecbf5e420f610a/analyzer/vulnerability-cache-processor-plugin/src/main/java/eu/fasten/analyzer/vulnerabilitycacheprocessorplugin/VulnerabilityCacheProcessorPlugin.java#L253

michelescarlato:
> > Could you please point me to some examples?
> 
> Here is an example that is quite similar to yours:
> https://github.com/fasten-project/fasten/blob/ce906724d2bfb85c8f6a943835ecbf5e420f610a/analyzer/vulnerability-cache-processor-plugin/src/main/java/eu/fasten/analyzer/vulnerabilitycacheprocessorplugin/VulnerabilityCacheProcessorPlugin.java#L253

Sorry @MihhailSokolov, I formulated my question wrongly.
Could you point me to an example of stitched call graph made in Java?
[This link](https://github.com/fasten-project/LCV-CM/blob/develop/src/LCV/json/benchmark-pycg-stitch/macro-out.json#L21148) points at a stitched call graph generated for python (please look starting from line 21148).

@cg122 Chushu, thank you for the suggestion, I think that also could be a possible solution, but it would probably be only java-specific (? ... still, I don't have clear of the state of Debian and Python).

I thought that operating directly on the stitched-call graph could have been a unique solution for all three use cases.
Your suggestion would imply recursive calls to the Fasten REST APIs to retrieve the full list of dependencies (which sounds much better than working on the full stitched call graph, as I was opting for Python). 

But this approach does not ensure that there will be license information detected by the `license detection` plugin in the database, I guess (that for the Maven use case starts to be fundamental since we cannot completely rely on the license information obtained from the pom.xml file, and Maven is not providing license information through API, on the contrary of pypi and Debian that they do).

MihhailSokolov:
> Could you point me to an example of stitched call graph made in Java?
> [This link](https://github.com/fasten-project/LCV-CM/blob/develop/src/LCV/json/benchmark-pycg-stitch/macro-out.json#L21148) points at a stitched call graph generated for python (please look starting from line 21148).

@michelescarlato 
I don't quite get why you would need a serialized (in JSON) DIrectedGraph for Java. Sorry, I don't have an example, because we never really serialize them, they are just used right away as in that example I mentioned earlier. 
But if you do need to serialize and deserialize DirectedGraphs for Java, you can use `eu.fasten.core.data.utils.DirectedHraphSerializer` and `eu.fasten.core.data.utils.DirectedHraphDeserializer` respectively.

michelescarlato:
@MihhailSokolov, my main concern regarding the License Compliance Verification, which is a separate process from the License Detection, is having all the information regarding the dependencies (and possibly their transitive dependencies) of an analyzed package.

As far as I understood, the stitched call graph can be considered a final step (or one of the final steps) of the whole pipeline. Do you agree with this information?

While for Debian and Python, I can rely on, at least, the license information declared at a package level by the packages developer, for Maven, I really need the license Detection phase to be completely performed (each dependency, and eventually their transitive dependencies).
I was assuming that just waiting for the creation of the stitched call graph was the best solution, mostly because looking at the Python stitched call graph, I could see information such as package versions and metadata.

e.g., this is taken from the line `21148` to `21154` of [this stitched call graph](https://github.com/fasten-project/LCV-CM/blob/develop/src/LCV/json/benchmark-pycg-stitch/macro-out.json#L21148) (generated by `pycg-stitch`)  :

```
"nodes":{
      "0":{
         "URI":"fasten://bcrypt$3.2.0/bcrypt/checkpw()",
         "metadata":{
            
         }
      },
```

The metadata field is supposed to be fulfilled with license information.
The package name `bcrypt` and its version `3.2.0` would be enough information to perform a compliance verification after collecting the same information for the other dependencies.
@vitsalis or @StefanosChaliasos, could you please confirm that inside the metadata field will end up license information (of course, after that, the License Detection will be performed)? 

To make more clear, the approach that I would take for the Python use case is:
1. Retrieve the license information regarding bcrypt-3.2.0 from here:
[https://pypi.org/pypi/bcrypt/3.2.0/json](https://pypi.org/pypi/bcrypt/3.2.0/json)
which is released under `Apache-2.0`, 

2. compare this license with the license information retrieved by the License Detection phase.
3. provide two different License compliance and verification assessments.
a. a short assessment considering only the license declared in pypi.org for each dependency, checking if there are compatibility issues with the license declared for the final package,
b. a verbose assessment including the license information collected during the license detection phase, using the scanning tool.

In the case of Maven, the main problem is that there are no APIs providing such information (point 1), and within the pom.xml, license information is not always included. This requires that each dependency be analyzed with the License Detector plugin to perform any compliance assessment.

But as far as I understand, your stitched call graph is completely different from the one produced by pycg-stitch. Right?

I now understand better your suggestion:
"The verification should happen on user request (for example, through the REST API). At this point, I would create a class that you can instantiate, call a method that would check license compliance for a certain artifact (and its dependency set), and return the result. Then this result can be used in the REST API or wherever else needed."

And my final question is:
" If there is no information regarding the license dependencies, the call to the License Compliance Verification (LCV) FASTEN REST API would then trigger the whole pipeline, right?"

Second question upon this reasoning:
"Potentially, how long could it take?"

MihhailSokolov:
@michelescarlato
Being able to stitch the whole dependency set requires that each one of the artifacts in that dependency set has gone through our ingestion pipeline. There is no point in the pipeline where we could guarantee that because the whole pipeline is based on processing only a single artifact at a time without any context.
This is exactly why the license verification should happen on user request and not related to the pipeline. It would be user's responsibility to make sure that all the artifacts from dependency set that he/she needs have been processed before making a call the some REST API endpoint that in turn would call your class for license verification. If any artifacts from the dependency set are not in the database when you want to stitch, you can stitch and perform further analysis, just give some kind of warning that not all artifacts from depset have been analysed (or something like that).
The main point is that it is impossible to guarantee that all artifacts in the dependency set have been processed, therefore this lies in the hands of user to make sure that he/she get the full result with all artifacts stitched.

michelescarlato:
Very clear answer @MihhailSokolov!
Thank you for the great clarification :)

proksch:
What is the state of this PR? Is there still active work performed or has this been superseded since and can be closed?

michelescarlato:
Hi Sebastian @proksch, thanks for the notification. Yes, currently, it is possible to close it since the License Compliance Verification (LCV) occurs outside of the pipeline.

PR #268, opened 2021-04-10
Setting callSiteOnly switch for Kafka Plugin (true) and CLI use (false)

MagielBruntink:
Changed the default of callSiteOnly in PartialCallGraph constructor to true, the Kafka Plugin uses true explicitly, and the CLI tool uses false.


ashkboos:
Thanks, @MagielBruntink for the PR and for pointing this out. I forgot to consider this switch in the opal plugin. Fortunately, the commit was after we restarted the deployment, so the CGs that we have been generating in the pipeline should be fine.

MagielBruntink:
Let’s merge? In our case our pipeline is running with the wrong value now and I want to restart the CG generation over the weekend. Just about 600 call graphs so no big problem.

PR #269, opened 2021-04-11
Sets output path of payload, so it can be stored to disk.

wzorgdrager:
OPAL can write its payload to disk (to prevent CG's which are too big for Kafka). In the metadata plugin, this output path was not set and therefore the write-payload-to-disk did not work. This PR resolves that issue.

### DIFF
--- a/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
+++ b/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
@@ -161,6 +161,9 @@ public class MetadataDBExtension implements KafkaPlugin, DBConnector {
                         restartTransaction = false;
                         logger.info("Saved the '" + revision + "' callgraph metadata "
                                 + "to the database with package version ID = " + id);
+
+                        // Set the output path for this call graph, so that it can be stored to disk if necessary.
+                        setOutputPath(callgraph);
                     }
                 });
             } catch (Exception expected) {


MihhailSokolov (approved):
LGTM! Just don't forget to update the deployment file of the metadata plugin to set the base output path to something like `/mnt/fasten/metadata-out`

wzorgdrager:
Thx for checking @MihhailSokolov. With your comment you mean I have to properly set the `po` and `pl` flags in the deployment?

MihhailSokolov:
> Thx for checking @MihhailSokolov. With your comment you mean I have to properly set the `po` and `pl` flags?  

Yes. I think `-pl` stays the same and you need to add a `-po` to set the base directory to which those messages will be written.

wzorgdrager:
> > Thx for checking @MihhailSokolov. With your comment you mean I have to properly set the `po` and `pl` flags?
> 
> Yes. I think `-pl` stays the same and you need to add a `-po` to set the base directory to which those messages will be written.

Thanks, but then I don't understand where this is done for OPAL then? Because there the `-po` is just `mnt/fasten`, I would expect the `/mvn` folder also somewhere. Do you know that?

MihhailSokolov:
> > > Thx for checking @MihhailSokolov. With your comment you mean I have to properly set the `po` and `pl` flags?
> > 
> > Yes. I think `-pl` stays the same and you need to add a `-po` to set the base directory to which those messages will be written.
> 
> Thanks, but then I don't understand where this is done for OPAL then? Because there the `-po` is just `mnt/fasten`, I would expect the `/mvn` folder also somewhere. Do you know that?

The `/mnt/fasten` is the base path. It is then appended with the actual path by using `setOutputPath()` inside the plugin. In the case of OPAL, the base path is `/mnt/fasten`, then when OPAL generates a CG for some coordinate `g:a:v`, it sets the output path to `/mvn/first_letter_of_g/a/g_a_v.json` (or something similar), and in the end the CG is stored under `/mnt/fasten/mvn/first_letter_of_g/a/g_a_v.json`. Same thing should happen to metadata. We need to set the base path to something like `/mnt/fasten/metadata-out` and the `setOutputPath` method will augment it to a full path for each message to store.

wzorgdrager:
> > > > Thx for checking @MihhailSokolov. With your comment you mean I have to properly set the `po` and `pl` flags?
> > > 
> > > 
> > > Yes. I think `-pl` stays the same and you need to add a `-po` to set the base directory to which those messages will be written.
> > 
> > 
> > Thanks, but then I don't understand where this is done for OPAL then? Because there the `-po` is just `mnt/fasten`, I would expect the `/mvn` folder also somewhere. Do you know that?
> 
> The `/mnt/fasten` is the base path. It is then appended with the actual path by using `setOutputPath()` inside the plugin. In the case of OPAL, the base path is `/mnt/fasten`, then when OPAL generates a CG for some coordinate `g:a:v`, it sets the output path to `/mvn/first_letter_of_g/a/g_a_v.json` (or something similar), and in the end the CG is stored under `/mnt/fasten/mvn/first_letter_of_g/a/g_a_v.json`. Same thing should happen to metadata. We need to set the base path to something like `/mnt/fasten/metadata-out` and the `setOutputPath` method will augment it to a full path for each message to store.

I understand! Thanks for clarifying :)

Issue #272, opened 2021-04-20
Include Pending Vulnerability Statements in the Vulnerability Consumer Injection Pipeline


elanzini:
The calls to the `lazy` ingestion made by the `vulnerability-consumer` are not well-handled by the current pipeline, which just proceeds with the rest of the statements to ingest. The idea is to include a `fasten.vulnerability.pending.out` kafka topic where those can be funnelled and later ingested once the package versions have been lazily ingested.

proksch:
This idea has been superseded by the pipeline extension.

PR #274, opened 2021-04-23
Pending topic in Vulnerability Consumer for lazy ingest

elanzini:
Takes care of #272, using a new `pending` topic to enqueue vulnerabilities that we were not able to inject and for which we used lazy ingestion. The logic is implemented by including a `KafkaProducer<String, String>` object in the `VulnerabilityConsumerExtension` class and we use it to publish to a new pending topic. 

To avoid recursive pending topics (e.g. pending consumer publishing to a `pending.pending` consumer), in the `setTopic` method, we check if we are already in a `pending` context by looking at the `topicName` to consume from passed by the FASTEN server (https://github.com/fasten-project/fasten/commit/14d7886837c7ae1969e3457a4a5cf13031916887#diff-6d444bb0169e676186af6b8a2a71a16fddbae23bf90ba36650fca11e6fa4ff8eR83).

### DIFF
--- a/analyzer/vulnerability-consumer/src/main/java/eu/fasten/analyzer/vulnerabilityconsumer/VulnerabilityConsumer.java
+++ b/analyzer/vulnerability-consumer/src/main/java/eu/fasten/analyzer/vulnerabilityconsumer/VulnerabilityConsumer.java
@@ -36,7 +36,10 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Arrays;
 import java.util.Optional;
+import java.util.concurrent.atomic.*;
 import java.util.stream.Collectors;
+
+import org.apache.kafka.clients.producer.*;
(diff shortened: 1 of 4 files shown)


proksch:
This change seems to be in conflict with the recent changes, both to the architecture of the pipeline and to the vulnerability components. Please update the code to the current state or reject and close this PR.

proksch:
@mir-am As you are the main driver behind the changes regarding prio topics and were also involved with Edoardos work... how do you assess this PR? Do you agree that it seems to be obsolete in light of the recent changes? It also seems that this change is incompatible with @MagielBruntink suggested enhancements...

I would propose to reject/close this PR.

proksch:
As it seems, this PR does not have a maintainer anymore. Given the concerns that I have expressed in the previous comment, I will close it.

Feel free to reopen in the future, should work on this PR resume.

PR #281, opened 2021-06-08
Split Callable Search + Update Vulnerability Object

elanzini:
On the `vulnerability-producer` side, `patches` are now extracting `original` and `new` line changed for every file to make the search more precise. The logic to handle the new format is implemented on the `consumer` as well. 

The `CWE` field of the `Vulnerability` class is also updated to match the format from the `producer` (fixing https://github.com/fasten-project/vulnerability-producer/issues/77).


MagielBruntink:
Hi Edo, sure I will review. I'm running the new consumer & producer now in a local setup; will let you know what happens :-)

MagielBruntink:
@elanzini So far it's running without surprises. It has obtained data from all sources and it now going through the patches. So far so good :-)

MagielBruntink (approved):
Looks good to me; both tools are running for several hours without issue.

PR #284, opened 2021-06-25
Refactorings on Java pipeline

ashkboos:
## Description
To improve the efficiency and maintainability of our code we made multiple major refactorings on Java CG generation, databases, and CG merging in this PR.

## Motivation and context
Previously we had two implementations of the merge algorithm to support DB-based and Local merging. In this PR we combined these two implementations to improve the maintainability. Also after finalizing the CG generation and merge algorithm, we found multiple optimizations and refactoring opportunities in metadata DB and graph DB schema that we address in this PR. We also introduced a new concept called call-site for partial CGs of the packages. We use this term to prevent having internal/external edges in RevisionCallGraphs and instead save the call-sites that are defined in the packages.
Here is the refactored metadata DB schema:

image

## Testing
The previous tests should pass after these refactorings or if needed some of the tests should be refactored.

## Task list  
- [x] Combining merge algorithms
- [x] Refactoring metadata DB schema
- [x] Refactoring graphDB
- [x] Refactoring ERCG
- [x] Making sure that the tests pass
- [x] Extracting and adding to the database information about type-level annotations


MagielBruntink:
Getting some class cast exceptions in our metadata DB plugin. Did the datamodel change? Were all JOOQ classes re-generated?

MihhailSokolov:
> Getting some class cast exceptions in our metadata DB plugin. Did the datamodel change? Were all JOOQ classes re-generated?

Yes, the model did change and all JOOQ were indeed re-generated. On this branch, I don't get any class cast exceptions. Maybe you are trying to use the class which has been removed or renamed?

MagielBruntink:
> > Getting some class cast exceptions in our metadata DB plugin. Did the datamodel change? Were all JOOQ classes re-generated?
> 
> Yes, the model did change and all JOOQ were indeed re-generated. On this branch, I don't get any class cast exceptions. Maybe you are trying to use the class which has been removed or renamed?

Problem is in our code - I will push some fixes to this branch.

MagielBruntink:
Fix is needed in the REST API callables endpoint though now that the created_at field is removed:
```
lower(?)) limit ? offset ?]; ERROR: column callables.created_at does not exist
Position: 145] with root cause
org.postgresql.util.PSQLException: ERROR: column callables.created_at does not exist
Position: 145
at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2284) ~[postgresql-9.4.1208.jar!/:9.4.1208]
at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2003) ~[postgresql-9.4.1208.jar!/:9.4.1208]
at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:200) ~[postgresql-9.4.1208.jar!/:9.4.1208]
at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:424) ~[postgresql-9.4.1208.jar!/:9.4.1208]
at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:161) ~[postgresql-9.4.1208.jar!/:9.4.1208]
at org.postgresql.jdbc.PgPreparedStatement.execute(PgPreparedStatement.java:155) ~[postgresql-9.4.1208.jar!/:9.4.1208]
at org.jooq.tools.jdbc.DefaultPreparedStatement.execute(DefaultPreparedStatement.java:214) ~[jooq-3.14.8.jar!/:na]
at org.jooq.impl.Tools.executeStatementAndGetFirstResultSet(Tools.java:4201) ~[jooq-3.14.8.jar!/:na]
at org.jooq.impl.AbstractResultQuery.execute(AbstractResultQuery.java:279) ~[jooq-3.14.8.jar!/:na]
at org.jooq.impl.AbstractQuery.execute(AbstractQuery.java:375) ~[jooq-3.14.8.jar!/:na]
```

MihhailSokolov:
> Fix is needed in the REST API callables endpoint though now that the created_at field is removed:
> 
> ```
> 2021-06-28 11:04:00.357 DEBUG 1 --- [io-8080-exec-10] o.s.web.servlet.DispatcherServlet : GET "/mvn/packages/log4j:log4j/1.2.16/callables/", parameters={}
> 
> 2021-06-28 11:04:00.358 DEBUG 1 --- [io-8080-exec-10] s.w.s.m.m.a.RequestMappingHandlerMapping : Mapped to eu.fasten.analyzer.restapiplugin.mvn.api.CallableApi#getPackageCallables(String, String, int, int, String, Long)
> 
> 2021-06-28 11:04:00.360 DEBUG 1 --- [io-8080-exec-10] o.s.web.servlet.DispatcherServlet : Failed to complete request: org.jooq.exception.DataAccessException: SQL [select "public"."callables"."id", "public"."callables"."module_id", "public"."callables"."fasten_uri", "public"."callables"."is_internal_call", "public"."callables"."created_at", "public"."callables"."line_start", "public"."callables"."line_end", "public"."callables"."metadata" from "public"."packages" join "public"."package_versions" on "public"."packages"."id" = "public"."package_versions"."package_id" join "public"."modules" on "public"."package_versions"."id" = "public"."modules"."package_version_id" join "public"."callables" on "public"."modules"."id" = "public"."callables"."module_id" where (true and lower("public"."packages"."package_name") = lower(?) and lower("public"."package_versions"."version") = lower(?)) limit ? offset ?]; ERROR: column callables.created_at does not exist
> 
> Position: 145
> 
> 2021-06-28 11:04:00.361 ERROR 1 --- [io-8080-exec-10] o.a.c.c.C.[.[.[/].[dispatcherServlet] : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is org.jooq.exception.DataAccessException: SQL [select "public"."callables"."id", "public"."callables"."module_id", "public"."callables"."fasten_uri", "public"."callables"."is_internal_call", "public"."callables"."created_at", "public"."callables"."line_start", "public"."callables"."line_end", "public"."callables"."metadata" from "public"."packages" join "public"."package_versions" on "public"."packages"."id" = "public"."package_versions"."package_id" join "public"."modules" on "public"."package_versions"."id" = "public"."modules"."package_version_id" join "public"."callables" on "public"."modules"."id" = "public"."callables"."module_id" where (true and lower("public"."packages"."package_name") = lower(?) and lower("public"."package_versions"."version") = lower(?)) limit ? offset ?]; ERROR: column callables.created_at does not exist
> 
> Position: 145] with root cause
> 
> org.postgresql.util.PSQLException: ERROR: column callables.created_at does not exist
> 
> Position: 145
> 
> at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2284) ~[postgresql-9.4.1208.jar!/:9.4.1208]
> 
> at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2003) ~[postgresql-9.4.1208.jar!/:9.4.1208]
> 
> at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:200) ~[postgresql-9.4.1208.jar!/:9.4.1208]
> 
> at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:424) ~[postgresql-9.4.1208.jar!/:9.4.1208]
> 
> at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:161) ~[postgresql-9.4.1208.jar!/:9.4.1208]
> 
> at org.postgresql.jdbc.PgPreparedStatement.execute(PgPreparedStatement.java:155) ~[postgresql-9.4.1208.jar!/:9.4.1208]
> 
> at org.jooq.tools.jdbc.DefaultPreparedStatement.execute(DefaultPreparedStatement.java:214) ~[jooq-3.14.8.jar!/:na]
> 
> at org.jooq.impl.Tools.executeStatementAndGetFirstResultSet(Tools.java:4201) ~[jooq-3.14.8.jar!/:na]
> 
> at org.jooq.impl.AbstractResultQuery.execute(AbstractResultQuery.java:279) ~[jooq-3.14.8.jar!/:na]
> 
> at org.jooq.impl.AbstractQuery.execute(AbstractQuery.java:375) ~[jooq-3.14.8.jar!/:na]
> 
> at org.jooq.impl.AbstractResultQuery.fetch(AbstractResultQuery.java:333) ~[jooq-3.14.8.jar!/:na]
> 
> at org.jooq.impl.SelectImpl.fetch(SelectImpl.java:2880) ~[jooq-3.14.8.jar!/:na]
> 
> at eu.fasten.core.data.metadatadb.MetadataDao.getPackageCallables(MetadataDao.java:1402) ~[core-0.0.1-SNAPSHOT.jar!/:na]
> 
> at eu.fasten.analyzer.restapiplugin.mvn.api.impl.CallableApiServiceImpl.getPackageCallables(CallableApiServiceImpl.java:43) ~[classes!/:na]
> 
> at eu.fasten.analyzer.restapiplugin.mvn.api.CallableApi.getPackageCallables(CallableApi.java:48) ~[classes!/:na]
> ```

Weird, I don't get this error when calling that endpoint. Are you sure your code is up-to-date with this branch?

MagielBruntink:
> Weird, I don't get this error when calling that endpoint. Are you sure your code is up-to-date with this branch?

My bad. I was actually running a different image for the REST API, which did indeed not have this branch's code. Sorry.

MagielBruntink:
Another observeration: with the new merger there is a difference between the number of nodes included in the allUris biMap and the nodes in the the merged directed graph:
```
var merger = new CGMerger(allCallGraphs);
var mergedGraph = merger.mergeAllDeps();
var mergedUris = merger.getAllUris();
mergedUris.size());          ### 17048 nodes
mergedGraph.nodes().size()); ### 13385 nodes
```
So, quite a few nodes less in the graph than in the URIs map. Is this expected? This is for JPacman merged with its dependencies.

MihhailSokolov:
> Another observeration: with the new merger there is a difference between the number of nodes included in the allUris biMap and the nodes in the merged directed graph:
> ```
> var merger = new CGMerger(allCallGraphs);
> var mergedGraph = merger.mergeAllDeps();
> var mergedUris = merger.getAllUris();
> mergedUris.size());          ### 17048 nodes
> mergedGraph.nodes().size()); ### 13385 nodes
> ```
> So, quite a few nodes less in the graph than in the URIs map. Is this expected? This is for JPacman merged with its dependencies.

I think @ashkboos can better answer that :)

ashkboos:
> Another observeration: with the new merger there is a difference between the number of nodes included in the allUris biMap and the nodes in the merged directed graph:
> 
> ```
> var merger = new CGMerger(allCallGraphs);
> var mergedGraph = merger.mergeAllDeps();
> var mergedUris = merger.getAllUris();
> mergedUris.size());          ### 17048 nodes
> mergedGraph.nodes().size()); ### 13385 nodes
> ```
> 
> So, quite a few nodes less in the graph than in the URIs map. Is this expected? This is for JPacman merged with its dependencies.

This is expected. mergedUris.size() contains all possible nodes in a dependency set whether or not they are reachable or Stitchable. MergedUris is built before stitching and is a superset of mergedGraph.nodes() and makes it possible to retrieve the Uris of all nodes.  mergedGraph.nodes() on the other hand is built while Stitching and only contains the stitched nodes.

MagielBruntink:
> This is expected. merger.getAllUris() contains all possible nodes in a dependency set whether or not they are reachable or Stitchable. MergedUris is built before stitching and is a superset of mergedGraph.nodes() and makes it possible to retrieve the Uris of all nodes. mergedGraph.nodes() on the other hand is built while Stitching and only contains the stitched nodes.

Ok. Reachable from where, though? What are the start/source nodes considered for the reachability?

What we do is merge a local code base to its dependencies, so we consider all nodes in the local code base to be start nodes. However now, we see that not all nodes included in the merged graph. Before this was different.

ashkboos:
> > This is expected. merger.getAllUris() contains all possible nodes in a dependency set whether or not they are reachable or Stitchable. MergedUris is built before stitching and is a superset of mergedGraph.nodes() and makes it possible to retrieve the Uris of all nodes. mergedGraph.nodes() on the other hand is built while Stitching and only contains the stitched nodes.
> 
> Ok. Reachable from where, though? What are the start/source nodes considered for the reachability?
> 
> What we do is merge a local code base to its dependencies, so we consider all nodes in the local code base to be start nodes. However now, we see that not all nodes included in the merged graph. Before this was different.

By reachable I mean they are calling or being called from other nodes that are present in the dependency set that we are merging. For example, method m() should be present in the result if any method in the dependency set is calling it. Also, m() is present if it is calling one or more methods in this dependency set.
Previously we had a flag to add also external nodes to the result which is now a bit different, maybe that's why you don't see some of the nodes that you would see before. However, it could also be a new bug that we introduced in our new implementations I appreciate it if you give us the details if you find any specific suspicious case.

MagielBruntink:
> > > This is expected. merger.getAllUris() contains all possible nodes in a dependency set whether or not they are reachable or Stitchable. MergedUris is built before stitching and is a superset of mergedGraph.nodes() and makes it possible to retrieve the Uris of all nodes. mergedGraph.nodes() on the other hand is built while Stitching and only contains the stitched nodes.
> > 
> > 
> > Ok. Reachable from where, though? What are the start/source nodes considered for the reachability?
> > What we do is merge a local code base to its dependencies, so we consider all nodes in the local code base to be start nodes. However now, we see that not all nodes included in the merged graph. Before this was different.
> 
> By reachable I mean they are calling or being called from other nodes that are present in the dependency set that we are merging. For example, method m() should be present in the result if any method in the dependency set is calling it. Also, m() is present if it is calling one or more methods in this dependency set.
> Previously we had a flag to add also external nodes to the result which is now a bit different, maybe that's why you don't see some of the nodes that you would see before. However, it could also be a new bug that we introduced in our new implementations I appreciate it if you give us the details if you find any specific suspicious case.

Thanks for the reply @ashkboos 
To be sure: a callable should be included in the merged graph if it either 
 - calls another callable, anywhere in the set of graphs provided to the merger, including it's own graph, or
 - is called by another callable from anywhere in the set of graphs provided to the merger, including it's own graph?
 
 I will indeed try to isolate an example that may be problematic (I'm not 100% sure).

ashkboos:
> To be sure: a callable should be included in the merged graph if it either
> 
> * calls another callable, anywhere in the set of graphs provided to the merger, including it's own graph, or
> * is called by another callable from anywhere in the set of graphs provided to the merger, including it's own graph?

Yes, that's correct.

MagielBruntink:
The issue may be somewhat different. I'm looking at a (non-merged) call graph for JPacman, and some local edges that I expect are simply not there. Example, in the attached zipped JSON, which was just now generated with latest OPAL., search for `BoardPanel.paint`. It is node id "4". In the [JPacman code](https://github.com/SERG-Delft/jpacman-framework/blob/6f703ade639d7d2c9a4152e748ad5696595f5226/src/main/java/nl/tudelft/jpacman/ui/BoardPanel.java#L67) it's clear that the paint method calls `render`, defined just below on line 80. `render` is an overloaded method, with node ids "5" and "6" in the call graph. There are however no edges between id "4", "5", and "6" in the call graph. Am I reading these graph files correctly?
[jpacman.zip](https://github.com/fasten-project/fasten/files/6728760/jpacman.zip)

In fact, `paint` calls more methods, also `game.getLevel().getBoard()`, `getSize()`, which also appear to be missing.

MagielBruntink:
@ashkboos thanks for looking into the merging results. Did you see my post above on missing edges in non-merged single call graphs eg for the jpacman example? That can also cause differences as nodes without edges are excluded from the merging, right?

ashkboos:
> @ashkboos thanks for looking into the merging results. Did you see my post above on missing edges in non-merged single call graphs eg for the jpacman example? That can also cause differences as nodes without edges are excluded from the merging, right?

@MagielBruntink Sorry for the late response, I was busy pinpointing the place that the node difference was happening(failing the test). 
I took a look at the jpacman JSON that we have on the develop branch, that one also misses this edge. Hence, it's not introduced in the implementations of this branch.
It also shouldn't be the reason for failing the test here. Because fails is happening only on CI not locally, and we are reading the same CGs from files in both places. I think the problem was while augmenting the results of multiple merges for dependencies. Because the number of edges was similar with only one node less on CI. For some reason, we were unifying two nodes here but were considering them two different nodes locally. I improved the implementation, hopefully, it won't happen again.
About the paint method in jpacman, however, I think it is because OPAL didn't add an edge there, hence we also did not add an edge while creating the ERCG. The way that OPAL works is that it starts from some entry points based on automatic configuration. We mostly use its library configuration in FASTEN, so it starts from methods that it finds as library APIs and iteratively builds the call graph from there by following the calls. Probably the paint methods did not appear in any of those paths.

MagielBruntink:
> About the paint method in jpacman, however, I think it is because OPAL didn't add an edge there, hence we also did not add an edge while creating the ERCG. The way that OPAL works is that it starts from some entry points based on automatic configuration. We mostly use its library configuration in FASTEN, so it starts from methods that it finds as library APIs and iteratively builds the call graph from there by following the calls. Probably the paint methods did not appear in any of those paths.

Somewhat off-topic for this branch then; but as we have this conversation now: would specifying the main class of an app (eg. jpacman) result in those edges be added?

ashkboos:
> @ashkboos thanks for looking into the merging results. Did you see my post above on missing edges in non-merged single call graphs eg for the jpacman example? That can also cause differences as nodes without edges are excluded from the merging, right?

ashkboos:
> > About the paint method in jpacman, however, I think it is because OPAL didn't add an edge there, hence we also did not add an edge while creating the ERCG. The way that OPAL works is that it starts from some entry points based on automatic configuration. We mostly use its library configuration in FASTEN, so it starts from methods that it finds as library APIs and iteratively builds the call graph from there by following the calls. Probably the paint methods did not appear in any of those paths.
> 
> Somewhat off-topic for this branch then; but as we have this conversation now: would specifying the main class of an app (eg. jpacman) result in those edges be added?

If the method is reachable from the main it should be added. However, note that you should also change the OPAL config from Library mode to Application mode. We have a logic to do that automatically when you pass the main method but in general, I find the configuration of OPAL a bit tricky especially when you run it as a library in Java(it's written in scala). That's why we added a huge config file to our repo to make sure we always get the same CG in all of our runs. You could also try to change the config file manually. 
It's also worth mentioning that for future improvements I have an idea in mind to completely get rid of the OPAL CG generation for creating ERCGs if possible. We can extract the call-sites directly from the bytecode. This way we will have all the possible nodes and edges. Selecting from them will be up to the user. However, this has some costs, it may increase the size of our CGs drastically and we have to reimplement a big part of our code. So we need to first investigate the feasibility of handling bigger CGs. Moreover, for the scientific part, we needed to compare our Stitching with OPAL to make sure our algorithm is correct. So the first goal was to get it accepted as an approach that is as correct as OPAL. After we get the acceptance I will look into the possibility of working with bytecode directly.

MagielBruntink:
@ashkboos When are you planning to merge?

ashkboos:
> @ashkboos When are you planning to merge?

We made a last-minute decision to add a feature that XWiki requested (annotation information for classes). The implementation is done, hopefully, we will merge tomorrow.

PR #287, opened 2021-07-15
Added environment variable for the FASTEN API URL

MagielBruntink:
Added environment variable for the FASTEN API URL and made the Vulnerability Consumer use it.

## Description
Added a FASTEN_API_URL environment variable and a default to the core constants.

With this change the Vulnerability Consumer can be configured with a environment variable FASTEN_API_URL to send its ingest requests to.

## Motivation and context
The central FASTEN KB was hard-coded to be the target of ingest requests; this is undesirable.


elanzini (approved):
lgtm

PR #294, opened 2021-07-22
Update default consumer topic for quality-analyser

amottier:
Set default quality-analyser consumer topic to `fasten.RapidPlugin.callable.out` in order to use the same value as k8s and Docker Compose deployments.


MagielBruntink (approved):
Yes, good change! Thanks.

PR #308, opened 2021-10-06
Add a graph to document pipeline deployment

amottier:
A documentation of the pipeline deployment created using draw.io (can be downloaded from: https://github.com/jgraph/drawio-desktop/releases/latest).
This should help to identify for each component of the pipeline:
- the Dockerfile that was used to build the Docker image deployed
- the name of the docker image deployed
- the location of the k8s yaml file
- the name of the service declared in Docker Compose

This also help to identify the name of Kafka topic and the flow of data between each pipeline components.

And finally this should help to locate the Docker images that are not deployed by k8s and/or Docker Compose.


amottier:
@MagielBruntink would you mind taking a quick look at this graph (a rendered version is available on https://github.com/fasten-project/fasten/wiki/Deployment) and let me know if you think any adjustment is required? Thanks a lot.

MagielBruntink:
Sorry @amottier I cannot find the time for your request. @cg122 could you please assist?

cg122:
I will have a look.

cg122:
Thanks for the good work @amottier. It is a bit challenge to read this diagram in my screen :glad

cg122:
Thanks for the nice work @amottier. A good overview of all possible deployments, although It is a bit challenge to read this big diagram in my screen. 

Some remarks:
- It would be better to have legends description. For example, there are three types of arrow. What do they mean? 
- For the consumed and produced Kafka topics, different icons should be better.
- The diagram is kind of mixing `k8s-deployments` and `fasten-docker-deployments` together. I would suggest to separate the two and make it more clear.
- Vulnerability deployment is updated for `fasten-docker-deployment`, `fasten-vulnerability-packages-listener` also listens to `fasten.POMAnalyzer.out` topic now.

amottier:
@cg122 I update the diagram with taking into account your review:
- I have added a legend
- For the consumed and produced Kafka topics I now use different colors
- I added the connection between POMAnalyzer and VulnerabilityPackagesKafkaPlugin

For now I keep only one diagram for both Docker Compose and k8s deployment as they should ultimately the same as we discuss in today call.

I updated the image on https://github.com/fasten-project/fasten/wiki/Deployment

Would you mind taking a look to this version and merge it if you think it address your feedback?

cg122:
Looks good to me.

PR #309, opened 2021-10-11
Don't loose externals when merging

tmortagne:
The FASTEN Maven plugin need to have also the external node in the graph to do binary compatibility analysis. Unfortunately those are currently wiped by CGMerge.

I did a first version where external node were never wiped and finally made this optional in case there was a need for this cleanup in other use cases.

### DIFF
--- a/core/src/main/java/eu/fasten/core/merge/CGMerger.java
+++ b/core/src/main/java/eu/fasten/core/merge/CGMerger.java
@@ -83,7 +83,7 @@ public class CGMerger {
     private List<Pair<DirectedGraph, ExtendedRevisionJavaCallGraph>> ercgDependencySet;
     private BiMap<Long, String> allUris;
 
-    private Map<String, Map<String, String>> externalUris = new HashMap<>();
+    private Map<String, Map<String, String>> externalUris;
 
     public BiMap<Long, String> getAllUris() {
         return this.allUris;
@@ -95,11 +95,24 @@ public class CGMerger {
      * @param dependencySet all artifacts present in a resolution
      */
     public CGMerger(final List<ExtendedRevisionJavaCallGraph> dependencySet) {
+        this(dependencySet, false);
+    }
+
+    /**
+     * Creates instance of callgraph merger.
+     *
+     * @param dependencySet all artifacts present in a resolution
+     * @param withExternals true if unresolved external calls should be kept in the generated graph
+     */
+    public CGMerger(final List<ExtendedRevisionJavaCallGraph> dependencySet, boolean withExternals) {
 
         final var UCH = createUniversalCHA(dependencySet);
         this.universalParents = UCH.getLeft();
         this.universalChildren = UCH.getRight();
         this.allUris = HashBiMap.create();
+        if (withExternals) {
+            this.externalUris = new HashMap<>();
+        }
         final var graphAndDict = getDirectedGraphsAndTypeDict(dependencySet);
         this.ercgDependencySet = graphAndDict.getLeft();
         this.typeDictionary = graphAndDict.getRight();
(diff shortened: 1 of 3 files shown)


tmortagne:
The Maven plugin does not just need the unresolved external calls. It needs a call graph with, among other things, the unresolved external calls.

I wrote this pull request to be the less impacting possible to the existing algo to avoid side effects (it's not even enabled by default, except for the map of external calls which should not be very expensive and could be optional too) but if you have performances improvements ideas involving refactoring the existing code don't hesitate :)

ashkboos:
Thanks for the clarification. Yes, I know that the Maven plugin needs a call graph and unresolved calls. My question is whether or not it is possible to extract the information that you need from the input and use it together with the call graph that the merger generates. Since the full URIs are available in the call graph it is also possible to match callables with any kind of other information in case you need both at the same time. Maybe it's easier if we set up a call and discuss this there, could we?

proksch:
I am not sure I get the motivation behind this change... wouldn't it be easier to do the binary compatibility analyses per dependency (for which, individually, you can find all details about the bytecode, incl. unresolved external calls, in the OPAL output)? What is the benefit of doing it on the merged graph?

tmortagne:
Like with the other analysis, to avoid reporting false positives you need to create a call graph of things really used from the point of view of the module, so you need a merged graph for it.

Also, I don't understand what you mean by "finding unresolved calls in the OPAL output". For a given JAR OPAL just extract all the calls internals and external, and you have no idea which of those external calls exist or not in other dependencies. Resolving external calls is actually the main job of CGMerger.

proksch:
What I mean is that the OPAL output still contains everything, including the bits and pieces that would be dropped later. But I see that it makes sense to leverage the resolution capabilities of the merger to understand which parts are actually being used.

@ashkboos What is  your counterargument against the proposal? If the extension is disabled by default, the overhead should be negligible... I think we just need to clearly communicate that the ids that are being generated on-the-fly for the unresolved calls cannot be found in the DB. Could we use very high numbers, to avoid an overlap with the DB? We might even be able to detect these high numbers in queries to the REST API and crash with an appropriate error message...

proksch (commented):
The changes itself look good to me.

proksch:
@ashkboos: Can we come to a data-driven conclusion? I have just reviewed the changes, and while I agree that complexity/maintainability is an important goal for us, the changes are rather small, so I am not too concerned. In addition, you claim that the overhead represents a performance problem, can you quantify how much slower an average CG generation will become?

ashkboos:
> @ashkboos: Can we come to a data-driven conclusion? I have just reviewed the changes, and while I agree that complexity/maintainability is an important goal for us, the changes are rather small, so I am not too concerned. In addition, you claim that the overhead represents a performance problem, can you quantify how much slower an average CG generation will become?

I just ran our efficiencyTest which merges a relatively big project and did not observe a substantial overhead. About external IDs, I would suggest that we generate negative ids for external nodes so that it's obvious to everyone that these are different IDs.

core/src/main/java/eu/fasten/core/merge/CGMerger.java
```diff
@@ -386,6 +425,34 @@ public DirectedGraph mergeWithCHA(final DirectedGraph callGraph, final GraphMeta
...
+            Map<String, String> typeMap = this.externalUris.get(type);
+            if (typeMap != null) {
+                String nodeURI = typeMap.get(arc.target.receiverSignature);
+
+                if (nodeURI != null) {
+                    // Find external node id
+                    Long target = this.allUris.inverse().get(nodeURI);
+                    if (target == null) {
+                        // Allocate a global id to the external node
+                        target = this.allUris.keySet().stream().max(Long::compareTo).orElse(0L) + 1;
```
proksch, line 443: After the discussion, it seems that this line should be replaced with logic that assigns negative ids, as the "regular" ids are all positive. This allows us to have a simple "counter" with the next negative number and would allow us to get rid of the search through all keys.

proksch (changes requested):

proksch:
I really like the idea of the negative ids. It seems that we have an agreement then.

@tmortagne Could you introduce this minor change and then rebase the PR onto develop once more to solve the conflicts that are existing now? After that, we can merge it into develop.

tmortagne:
> @tmortagne Could you introduce this minor change and then rebase the PR onto develop once more to solve the conflicts that are existing now? After that, we can merge it into develop.

Sure.

tmortagne:
Should be ready now.

ashkboos (approved):

proksch (changes requested):
The changes look good to me. Could you rebase this PR on develop (not just merge)? A linear history makes changes sooo much easier to track.

tmortagne:
Not sure what you mean, I had to merge because of conflicts in CGMerger.

proksch:
What I mean is use `git rebase` instead of `git merge` ^^

Yes, if there are conflicts, you also have to solve them there, but then you end up not needing a merge commit and with a linear commit history.

tmortagne:
I usually avoid anything that require using `-f` when pushing on github but here is it.

proksch (approved):

proksch:
Thanks, highly appreciated! As long as you are working alone on a feature branch, force pushing is not a problem. It only becomes problematic, when you change history that somebody else has checked out. :)

PR #315, opened 2021-10-13
Adding normal/priority message processing to the FASTEN server

mir-am:
## Description
- Plug-ins now can now accept a number of Kafka topics to consume from, instead of only one topic.
- Given a list of input Kafka topics, the FASTEN server can detect normal and priority records and it processes the priority records first if any.

## Motivation and context
Previously, we had to maintain two different deployments for normal and priority topics. Moreover, the priority records did not have precedence over normal records when processing.

## Testing
On the servers, we need to run the plug-ins and make sure both priority and normal records are processed.

## Task list  
- [x] Adapt FASTEN plug-ins to accept a number of input topics.
- [x] In the FASTEN server, first process priority records (if any) and then normal records
- [x] Testing with a dummy plugin

## Additional context
To differentiate between normal and priority topics, the keyword `priority` should be included in the name of the input Kafka topics.


proksch:
I am really sorry, but this pull request is too large to review... it contains many unrelated changes in a lot of files. Looking at the history, it seems that there have been multiple merges back and forth between the "develop" branch and the feature branch, this makes it impossible to review the changes of the individual feature in isolation (e.g., there are several commits by roman on the vulnerability path part, which are COMPLETELY unrelated). In the future, please use rebasing, if you need access to changes that are only available on the base branch and esp. *before opening a PR*.

How do we handle this? I am tempted to reject this PR, but I am also not sure how to properly fix this situation here any more.

server/src/main/java/eu/fasten/server/FastenServer.java
```diff
@@ -219,77 +241,70 @@ public void run() {
...
-                DSLContext dbContext = null;
-                try {
-                    dbContext = getDSLContext(dbUrls.get("mvn"));
-                } catch (SQLException ex) {
-                    logger.error("Couldn't set {} DB connection for plug-in {}\n{}",
-                            dbUrls.get("mvn"), p.getClass().getSimpleName(), ex.getStackTrace());
-                }
-                p.loadGraphResolver(dbContext, depGraphPath);
-            } else {
+            if (dbUrls == null || depGraphPath == null) {
```
proksch, line 301: Shouldn't it be necessary to end this if with a 'return' statement to prevent an NPE right after?

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -228,7 +254,7 @@ public void handleConsuming() {
      * <p>
      * This strategy provides at-least-once semantics.
      */
-    public void processRecord(ConsumerRecord<String, String> record, Long consumeTimestamp) {
+    public void processRecord(ConsumerRecord<String, String> record, Long consumeTimestamp, boolean priority) {
```
proksch, line 257: Please rename to 'isPriority'

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -174,39 +185,54 @@ public void stop() {
...
     public void handleConsuming() {
-        ConsumerRecords<String, String> records = connection.poll(Duration.ofSeconds(1));
-        Long consumeTimestamp = System.currentTimeMillis() / 1000L;
+        boolean hasPrio = false;
+
+        if (!prioTopics.isEmpty()) {
+            ConsumerRecords<String, String> prioRecords = connPrio.poll(this.pollTimeout);
+            for (var r : prioRecords) {
+                logger.info("Read priority message offset " + r.offset() + " from partition " + r.partition() + ".");
+                processRecord(r, System.currentTimeMillis() / 1000L, true);
```
proksch, line 194: Use an enum instead of a boolean flag.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -174,39 +185,54 @@ public void stop() {
...
+                hasPrio = true;
+                logger.info("Successfully processed priority message offset " + r.offset() + " from partition " + r.partition() + ".");
+                // TODO: Keep a list of processed priority messages like normal ones
+            }
+            doCommitSync(true);
```
proksch, line 199: use enum here as well.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -174,39 +185,54 @@ public void stop() {
...
-        ConsumerRecords<String, String> records = connection.poll(Duration.ofSeconds(1));
-        Long consumeTimestamp = System.currentTimeMillis() / 1000L;
+        boolean hasPrio = false;
+
+        if (!prioTopics.isEmpty()) {
+            ConsumerRecords<String, String> prioRecords = connPrio.poll(this.pollTimeout);
+            for (var r : prioRecords) {
+                logger.info("Read priority message offset " + r.offset() + " from partition " + r.partition() + ".");
+                processRecord(r, System.currentTimeMillis() / 1000L, true);
+                hasPrio = true;
```
proksch, line 195: Rename to 'hasConsumedPriorityRecord'

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
proksch, line 257: use enum instead of boolean

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -258,15 +284,24 @@ public void processRecord(ConsumerRecord<String, String> record, Long consumeTim
...
      *
      * @param input input message [can be null]
      */
-    public void handleProducing(String input, long consumeTimestamp) {
+    public void handleProducing(String input, long consumeTimestamp, boolean priority) {
+        String outputTopicName;
+        String errorTopicName;
+        if (priority) {
+            outputTopicName = String.format("fasten.%s.priority.out", outputTopic);
+            errorTopicName = String.format("fasten.%s.priority.err", outputTopic);
```
proksch, line 300: Use a joint error topic for both prio and normal processing.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -399,13 +432,17 @@ private String getStdErrMsg(String input, Throwable pluginError, long consumeTim
      * This is a synchronous commits and will block until either the commit succeeds
      * or an unrecoverable error is encountered.
      */
-    private void doCommitSync() {
+    private void doCommitSync(boolean priority) {
```
proksch, line 435: use enum

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -432,22 +469,22 @@ private void skipPartitionOffsets() {
             return;
         }
         // Note that this assumes that the consumer is subscribed to one topic only
-        for (PartitionInfo p : this.connection.partitionsFor(topics.get(0))) {
+        for (PartitionInfo p : this.connNorm.partitionsFor(topics.get(0))) {
```
proksch, line 472: Either this method is incomplete (i.e., also add handling for connPrio), or it is not used at all, in which case, we should remove it altogether.

proksch (commented):
Initial pass through monstrous pull request.

server/src/main/java/eu/fasten/server/FastenServer.java
mir-am: As we discussed, I've added RuntimeException to prevent NPE

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Done!

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Created `Enum {NORMAL, PRIORITY}`

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Done!

server/src/main/java/eu/fasten/server/FastenServer.java
```diff
@@ -299,9 +298,9 @@ private void registerDBConnections(List<DBConnector> dbPlugins) {
     private void loadDependencyGraphResolvers(List<DependencyGraphUser> plugins) {
         plugins.forEach(p -> {
             if (dbUrls == null || depGraphPath == null) {
-                logger.error("Couldn't load dependency graph. Make sure that you have "
-                    + "provided a valid DB URL, username, password, "
-                    + "and a path to the serialized dependency graph.");
+                throw new RuntimeException("Couldn't load dependency graph. Make sure that you have "
```
proksch, line 301: Could you use an IllegalArgumentException instead? RuntimeException is a very general type, the error should communicate the problem.

analyzer/vulnerability-statements-processor/src/test/java/eu/fasten/vulnerabilitystatementsprocessor/VulnerabilityStatementsProcessorTest.java
```diff
@@ -30,13 +30,7 @@
...
 
 import java.text.ParseException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.Map;
+import java.util.*;
```
proksch, line 33: Could you please check which IntelliJ auto-formatting rule is responsible for this change? Please adopt this setting it and make it available for everybody on the repository level, this is usually easily possible in Jetbrains producs by placing the config in the right place. (See https://www.jetbrains.com/help/idea/configuring-code-style.html and look for "project level).

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -90,7 +92,11 @@ public FastenKafkaPlugin(boolean enableKafka, Properties consumerProperties, Pro
         this.plugin = plugin;
 
         if (enableKafka) {
-            this.connection = new KafkaConsumer<>(consumerProperties);
+            this.connNorm = new KafkaConsumer<>(consumerProperties);
+            // For priority connection, the client should be different from the normal one
+            Properties consumerPrioProperties = (Properties) SerializationUtils.clone(consumerProperties);
```
proksch, line 96: Please factor out a helper method `cloneWithNewId(p: Properties, prefix: String) : Properties` that clones he properties and sets a new id by appending a provided prefix to the old one.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -144,14 +154,15 @@ public void run() {
                 if (plugin.consumeTopic().isPresent()) {
                     handleConsuming();
                 } else {
-                    doCommitSync();
-                    handleProducing(null, System.currentTimeMillis() / 1000L);
+                    doCommitSync(false);
```
proksch: This is not yet using the new enum.

proksch (commented):
I have several follow-up comments.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Already addressed in 062910f

server/src/main/java/eu/fasten/server/FastenServer.java
mir-am: Done!

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: I've created a private method to do this. Thanks for the comment.

analyzer/vulnerability-statements-processor/src/test/java/eu/fasten/vulnerabilitystatementsprocessor/VulnerabilityStatementsProcessorTest.java
mir-am: Addressed in b06f236

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -129,6 +127,12 @@ public FastenKafkaPlugin(boolean enableKafka, Properties consumerProperties, Pro
         logger.debug("Constructed a Kafka plugin for " + plugin.getClass().getCanonicalName());
     }
 
+    private Properties cloneKafkaPropWithNewClientId(Properties consumerProperties, String suffix) {
+        Properties consumerPrioProperties = (Properties) SerializationUtils.clone(consumerProperties);
```
proksch, line 139: I did not realize this before... Why do we ned to de-/serialize the whole properties object for cloning? This introduces quite a bit of overhead. If Kafka indeed does have no way of simply copying the options, I would propose that we try move the properties creation into a static helper method that can be used from both the `FastenServer` and from this location.

mir-am: AFAIK, Kafka doesn't have a way to copy properties.
I have addressed this in d69255b by creating a static method.

server/src/main/java/eu/fasten/server/connectors/KafkaConnector.java
```diff
@@ -96,4 +97,10 @@ public static Properties kafkaProducerProperties(List<String> serverAddresses, S
                 "50000000"); //Set produce size to 50MB.
         return p;
     }
+
+    public static Properties cloneKafkaConsumerPropWithNewClientId(Properties consumerProperties, String suffix) {
+        Properties consumerPrioProperties = (Properties) SerializationUtils.clone(consumerProperties);
```
proksch, line 102: This still uses a serialization call and a cast. Please get rid of that.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -101,7 +100,7 @@ public FastenKafkaPlugin(boolean enableKafka, Properties consumerProperties, Pro
         if (enableKafka) {
             this.connNorm = new KafkaConsumer<>(consumerProperties);
             // For priority connection, the client should be different from the normal one
-            this.connPrio = new KafkaConsumer<>(cloneKafkaPropWithNewClientId(consumerProperties, "_priority"));
+            this.connPrio = new KafkaConsumer<>(KafkaConnector.cloneKafkaConsumerPropWithNewClientId(consumerProperties, "_priority"));
```
proksch, line 103: Why has this been extracted to another class now? I think the original idea was to factor-out the property creation and instantiate both property objects together in the `FastenServer` (or was it `Fasten KafkaPlugin`?), instead of passing one around and creating the other one somewhere else..

server/src/main/java/eu/fasten/server/connectors/KafkaConnector.java
mir-am: Removed the method!

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Done!

mir-am:
I close this PR as I don't want to push 100 commits (probably due to the incorrect use of `git rebase`)! So I created a new PR #330.

amottier:
@mir-am is it ok to remove https://github.com/fasten-project/fasten/tree/server-normal-prio branch ?

mir-am:
> @mir-am is it ok to remove https://github.com/fasten-project/fasten/tree/server-normal-prio branch ?

Yes. it's okay. The PR is already merged into develop.

PR #318, opened 2021-10-26
Optimizing the output of POMAnalyzer

mir-am:
## Description
The dependency data is removed from the output message of POMAnlyzer.

## Motivation and context
Currently, the output messages of POM Analyzer can be large due to the dependency information. However, this information is not used by downstream plug-ins via Kafka and also the dependencies data are stored in the metadata DB.

## Testing
We need to make sure that the upstream plug-ins will not break if the dependency info is absent for the output messages of POM Analyzer.

## Task list  
- [x] Removed the dependency data from the output JSON of POMAnalyzer

### DIFF
--- a/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java
+++ b/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java
@@ -21,16 +21,10 @@ package eu.fasten.analyzer.pomanalyzer;
 import eu.fasten.analyzer.pomanalyzer.pom.DataExtractor;
 import eu.fasten.core.data.Constants;
 import eu.fasten.core.data.metadatadb.MetadataDao;
-import eu.fasten.core.maven.utils.MavenUtilities;
 import eu.fasten.core.maven.data.DependencyData;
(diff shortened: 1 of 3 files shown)


mir-am:
@MagielBruntink @michelescarlato 
What is the verdict here?

MagielBruntink:
LGTM, but we still need to hear from @michelescarlato

michelescarlato:
> It makes sense to not include all those data in Kafka messages that has also been stored in a database table and as a JSON field.
> 
> @michelescarlato Do you plugins not use the POM Analyzer messages for further processing?

@marcomicera do our plugins use it?

proksch:
@mir-am if this PR is still valid, please rebase it onto develop, so we fast-forward merge it into develop. If not, then please close it.

mir-am:
I would like to close this PR since we have not heard back from the EC team. So I am not sure whether merging this PR will be 100% safe and without bad side effects for other partners.

marcomicera:
I confirm that [`license-detector`](https://github.com/fasten-project/fasten/tree/endocode/compliancePlugin/analyzer/license-detector) and [`license-feeder`](https://github.com/fasten-project/fasten/tree/endocode/compliancePlugin/analyzer/license-feeder) do not consume `fasten.POMAnalyzer.out` records.

PR #322, opened 2021-11-03
Fix REST API application and container not exiting on error

MagielBruntink:
## Description
Currently the REST API application does not properly exit if it encounters an error like not connecting to the DB. The problem is with Spring Boot as described here https://github.com/spring-projects/spring-boot/issues/27049

I upgraded the Spring Boot version and added the suggested entry to the application.properties file. Now the REST API application and its container properly restart and retry if there is an error connecting.


mir-am (approved):
LGTM!
Thanks for the fix.

PR #324, opened 2021-11-07
Add C & Python Dependents Resolution

gdrosos:
## Description
This Pull Request aims to add the implementation of the Dependents Resolution for both C and Python. The Resolver is located in the FASTEN core directory under the `/dependents/` subdirectory. It follows exactly the same implementation as the Maven Resolver and it can be used either as a standalone tool or be deployed e.g. through a REST-API.  The tool can be deployed for either C or Python, depending on two  inputs given  differing for each language, namely the metadata database url or the serialized dependency graph path.

## Testing
The Testing follows the same manner as that of Maven Resolver, and more Tests may be added in the future.

## Additional Context
* Some of the methods may change or be extended based on the REST-API implementation that is going to use the Resolver.
* Currently the resolver returns for each package version (located in the metadata database) all its dependent's package versions. This, -if required- may change to returning only the latest version of each dependent package.


amottier:
@gdrosos is this pull request ready for review? I'm not probably the best person to review it but I can at least take a first look at the code if it can be useful.

gdrosos:
Yes, this Pull Request is ready for review.

mir-am:
> Yes, this Pull Request is ready for review.

I can take a look at the code. However, I'm not very familiar with the dependency resolvers.

gdrosos:
Well, the implementation is exactly the same as that of Maven Resolver, but it is generalized in order to be used for C and Python dependents only, independently of the Maven Resolver, which has more utilities than just Dependents Resolution.

core/src/main/java/eu/fasten/core/dependents/data/Dependency.java
```diff
@@ -0,0 +1,483 @@
...
+    public final String scope;
+    public final boolean optional;
+    public final String type;
+    public final String classifier;
+
+    /**
+     * Valid dependency scopes. Defined by maven.
+     * Learn more: http://maven.apache.org/pom.html
+     */
+    public static final String[] SCOPES = {
```
mir-am, line 48: These scopes are for Maven. Do you think it is unnecessary here?

core/src/main/java/eu/fasten/core/dependents/data/Dependency.java
```diff
@@ -0,0 +1,483 @@
...
+            "compile",
+            "provided",
+            "runtime",
+            "test",
+            "system",
+            "import"
+    };
+
+    /**
+     * Constructor for Dependency object.
```
mir-am, line 58: Again, this is for maven. The method parameters may not be relevant for Python & C.

core/src/main/java/eu/fasten/core/dependents/utils/DependencyGraphUtilities.java
```diff
@@ -0,0 +1,190 @@
...
+        kryo.register(Class.forName("eu.fasten.core.dependents.data.Dependency$Exclusion"));
+        kryo.register(java.sql.Timestamp.class);
+        kryo.register(java.util.ArrayList.class);
+        kryo.register(Class.forName("java.util.Collections$UnmodifiableSet"));
+        kryo.register(org.apache.maven.artifact.versioning.DefaultArtifactVersion.class,
+                new DefaultPackageVersionSerializer());
+        kryo.register(org.apache.maven.artifact.versioning.ComparableVersion.class);
+        kryo.register(Class.forName("org.apache.maven.artifact.versioning.ComparableVersion$ListItem"));
+        kryo.register(Class.forName("org.apache.maven.artifact.versioning.ComparableVersion$IntItem"));
+        kryo.register(Class.forName("org.apache.maven.artifact.versioning.ComparableVersion$StringItem"));
```
mir-am, line 103: These are for Maven! What should it be for Python & C?

core/src/test/java/eu/fasten/core/dependents/data/DependencyTest.java
```diff
@@ -0,0 +1,160 @@
...
+        var expected = new Dependency("package", "4.11");
+        var json = expected.toJSON();
+        var actual = Dependency.fromJSON(json);
+        assertEquals(expected, expected);
+    }
+
+
+    @Test
+    public void equalsTest() {
+        Assertions.assertEquals(new Dependency("package", "4.12"), new Dependency("package", "4.12", new ArrayList<>(), "compile", false, "jar", ""));
```
mir-am, line 41: This test seems to be for Java. For example, it contains jar and compile.

core/src/main/java/eu/fasten/core/dependents/data/Dependency.java
gdrosos: Well, indeed the scope attribute is unnecessary right now, so I will delete it

core/src/main/java/eu/fasten/core/dependents/data/Dependency.java
gdrosos: You are right,  the only Dependency information currently retrieved from the C & Python Database is the package Name and the Version Constraints, so I will remove the unnecessary Dependency & Dependnecy Edge attributes.

core/src/test/java/eu/fasten/core/dependents/data/DependencyTest.java
gdrosos: Yes, right, I removed this test since Dependency Object now has only  two attributes

core/src/main/java/eu/fasten/core/dependents/utils/DependencyGraphUtilities.java
gdrosos: According to the [revision call graph format documentation](https://github.com/fasten-project/fasten/wiki/Revision-Call-Graph-format), the Python and C dependency Version Constraints follow the maven format. Therefore, we used the same class employed in Maven Resolver(```org.apache.maven.artifact.versioning.DefaultArtifactVersion```) as the type of Python and C package versions and also  used the method compareTo to make comparsions between them. Thus, on serialization, we need to register the same corresponding Classes, jsut like Maven Resolver.

gdrosos:
Hello @mir-am, thank you very much for thoroughly reviewing my PR. I tried addressing your comments on https://github.com/fasten-project/fasten/pull/324/commits/ab2efa8c3aeff158962e684a18db5e6677995fc8 . Regarding serialization, since package versions are of type ```org.apache.maven.artifact.versioning.DefaultArtifactVersion```, I followed the same serialization  procedure as that of MavenResolver (see https://github.com/fasten-project/fasten/pull/324#discussion_r752602911). Please let me know if anything else needs to be addressed before merging, so that I can move on with the Pull Request of the REST-API plugin extension.

mir-am:
> Hello @mir-am, thank you very much for thoroughly reviewing my PR. I tried addressing your comments on [ab2efa8](https://github.com/fasten-project/fasten/commit/ab2efa8c3aeff158962e684a18db5e6677995fc8) . Regarding serialization, since package versions are of type `org.apache.maven.artifact.versioning.DefaultArtifactVersion`, I followed the same serialization procedure as that of MavenResolver (see [#324 (comment)](https://github.com/fasten-project/fasten/pull/324#discussion_r752602911)). Please let me know if anything else needs to be addressed before merging, so that I can move on with the Pull Request of the REST-API plugin extension.

Thanks for addressing the comments. I will look at the changes again and let you know about the final verdict.
Also, it would be good if @amottier could take a look at the PR as he volunteered.

amottier:
Yes sure I can do a review of the code.

But before I do that I guess I would need a few clarification regarding the goal of this code in the overall Fasten pipeline.

Maybe my most important question is: does this code run as part of the Fasten pipeline (for example by plugins who scrap central package repositories) or is it triggered when the server receive a REST API query?

Currently my understanding is that this code is execute when querying the REST API to get information about a given package dependencies. To get those information the code actually query the metadata database. Information are in the metadata database thanks to plugin such as POM analyzer who use information received from Maven Central and Maven code to analyze the package dependencies and store the result in the database.

If my understanding is more or less correct I'm curious to understand why we have in the code a function named `findMatchingRevisions`? Because based on my understanding resolving the actual dependency version to use is the job of the tool such as Maven, Pip or APT so I would expect to have this kind of processing already done for example by POM analyzer plugin.

Maybe I'm totally wrong and maybe a quick call can help me better understand the architecture and should allow me to do a better review of the code.

gdrosos:
Thanks @amottier, well, currently  we have deployed 4 rest-apis throught flask for the stitching of C & Python Call Graphs and also for the dependency resolution.  We also need to have two endpoints for the dependents resolution. Those endpoints will be deployed throught extending the rest-api plugin and have the form of: ```/pypi/packages/shap/0.34.0/resolve/dependents?transitive=false``` and ```/debian/packages/shap/0.34.0/resolve/dependents?transitive=true```.

In order to create the two endpoints for the dependents resolution, we need to be able to construct the C & Python Global Dependency Graph from the dependency information from the metadata db, just like Java . This information will be stored to the Database throught the Metadata Plugin. Unfortunately, because I encountered some issues with the plugin and it would require some time to fix it, I decided to create a simple python script consuming the call graphs and fastly populating the db, so that I can move on with the implementation of the endpoints because it was more urgent. After deploying the endpoints, I am planing of fixing the MetadataDB extension and and populate the db within the plugin so that the whole procedure is included in the pipeline.

 Regarding the ```findMatchingRevisions```,  each Revision has Dependency information in the form of package with some Version Constraints. Those constraints  are stored in the Call Graphs and consequently in the Metadata DB in a [specific format](https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN402). Examples of a constraint can also be found in the [Wiki](https://github.com/fasten-project/fasten/wiki/Revision-Call-Graph-format). When we need to resolve the dependency of a Revision in order to construct the Dependency Graph, we need to select only the Revisions which exist on the MetadataDB whose Version is included within the Version Constraints that this revision has. To adduce an example, if a package A with version 1.0 depends on package B with version constraints [1.0-.2.0] then for all the Versions of the package B that exist on the Database and thus on the Global Dependency Graph, we need to add as an edge only the Revisions with Version between  [1.0-.2.0]. For example Revision B:1.4 is  considered a matching revision whereas B:2.7 not.

gdrosos:
If you want I can submit a draft Pull Request with the Rest-API plugin extension so that you can have a better view of how I am using the Dependency Graph generated by this Pull Request to create the endpoints.

amottier (commented):
Overall the PR looks good to me, I had only very minors comments.
I managed to successfully run the tests.
For it is ok to be merged.

proksch:
The PR has been approved by @amottier beginning of December, unfortunately, some conflicts emerged since. I would like to ask you to solve these conflicts (preferably in a rebase ;)) and merge the changes back into `develop`... it seems that also PR #351 depends on this change, this answer applies there as well.

core/src/main/java/eu/fasten/core/dependents/DependencyGraphBuilder.java
```diff
@@ -0,0 +1,195 @@
...
+        return revisions.stream().filter(r -> {
+            for (var constraint : constraints) {
+                if ((constraint.toString().startsWith("[") || constraint.toString().startsWith("("))
+                        && (constraint.toString().endsWith("]") || constraint.toString().endsWith(")"))) {
+                    if (checkVersionLowerBound(constraint, r.version) &&
+                            checkVersionUpperBound(constraint, r.version)) {
+                        return true;
+                    }
+                } else {
+                    if (constraint.lowerBound.equals(constraint.upperBound) &&
```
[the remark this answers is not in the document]
gdrosos: Hello @amottier, sorry for the late response. The purpose of the findMatchingRevisions method is to find revisions that match  the version constraints of each dependency. The checks regarding a valid Version constraint are performed in the Dependency.VersionConstraint Class. Each Version Constraint has a lower & upper Bound attributes and two isLowerHardRequirement & isUpperHardRequirement binary values. This method https://github.com/fasten-project/fasten/blob/0c1b112713e0b05c788957920f1c9962b19a3464/core/src/main/java/eu/fasten/core/dependents/data/Dependency.java#L239-L264 ensures that always either the Version Constraint will include parenthesis or brackets or it will be a specific version constraint. So the case where a constraint neither  does include parenthesis or brackets nor has a lower bound equal to upper bound will not happen.

.gitignore
```diff
@@ -190,4 +190,11 @@ modules.xml
 
 # End of https://www.gitignore.io/api/intellij
 
+!/analyzer/javacg-wala/src/test/resources/*.jar
+!/analyzer/javacg-opal/lib/**/*.jar
+!/analyzer/javacg-opal/src/test/resources/*.class
```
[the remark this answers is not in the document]
gdrosos: Thank you very much for pointing this out, i removed those extensions, as they were added mistakenly.

gdrosos:
> The PR has been approved by @amottier beginning of December, unfortunately, some conflicts emerged since. I would like to ask you to solve these conflicts (preferably in a rebase ;)) and merge the changes back into `develop`... it seems that also PR #351 depends on this change, this answer applies there as well.

Hello @proksch, it seems that this branch has no conflicts with the base branch, so can I rebase and merge this PR?

PR #329, opened 2021-11-09
Callable index merger improvement

ashkboos:
## Description
This PR uses the callable index to create the type dictionary in the merger.

## Motivation and context
It is important to use the callable index as much as possible instead of using metadata DB for better performance.

## Testing
It passes all the unit tests, also it is tested on the server manually to see if it produces the results.


core/src/main/java/eu/fasten/core/merge/CGMerger.java
```diff
@@ -493,32 +495,51 @@ private void resolve(final Set<LongLongPair> edges,
...
-                    newestSet.add(callable.value2().longValue());
-                    type.merge(node.signature, newestSet, (old, newest) -> {
-                        old.addAll(newest);
-                        return old;
-                    });
-                });
+        int noCGCounter = 0, noMetadaCounter = 0;
+        for (Long dependencyId : dependenciesIds) {
+            DirectedGraph cg;
+            try {
```
proksch, line 508: Maybe moving this to a helper method could help shorten this algorithm and make it more understandable

core/src/main/java/eu/fasten/core/merge/CGMerger.java
```diff
@@ -493,32 +495,51 @@ private void resolve(final Set<LongLongPair> edges,
...
+                metadata = rocksDao.getGraphMetadata(dependencyId, cg);
+            } catch (RocksDBException e) {
+                throw new RuntimeException("An exception occurred retrieving metadata from rocks " +
+                    "DB", e);
+            }
+            if (metadata == null) {
+                noMetadaCounter++;
+                continue;
+            }
+            logger.info("For {} dependencies failed to retrieve {} graph data and {} metadata " +
```
proksch, line 528: I think this logger statement should be placed after the loop.

proksch (changes requested):
I have two minor reservations before we can accept this change.

proksch (approved):

PR #330, opened 2021-11-12
Adding normal/priority message processing to the FASTEN server

mir-am:
## Description
- Plug-ins now can now accept a number of Kafka topics to consume from, instead of only one topic.
- Given a list of input Kafka topics, the FASTEN server can detect normal and priority records and it processes the priority records first if any.

## Motivation and context
Previously, we had to maintain two different deployments for normal and priority topics. Moreover, the priority records did not have precedence over normal records when processing.

## Testing
On the servers, we need to run the plug-ins and make sure both priority and normal records are processed.

## Task list  
- [x] Adapt FASTEN plug-ins to accept a number of input topics.
- [x] In the FASTEN server, first process priority records (if any) and then normal records
- [x] Testing with a dummy plugin

## Additional context
To differentiate between normal and priority topics, the keyword `priority` should be included in the name of the input Kafka topics.


proksch:
I see that this is a resubmission of #315 after that one has been closed due to Git problems. Apart from the rebasing issue, all other comments have been fixed in #315. Have there been any other changes that need reviewing or can we just accept this PR?

mir-am:
> I see that this is a resubmission of #315 after that one has been closed due to Git problems. Apart from the rebasing issue, all other comments have been fixed in #315. Have there been any other changes that need reviewing or can we just accept this PR?

I think we can merge this PR.

PR #331, opened 2021-11-16
Handling RuntimeExceptions and UnrecoverableErrors in the FASTEN server

mir-am:
## Description
- A custom exception, `UnrecoverableError`, is defined for rare cases where we deliberately want to crash a plug-in and cause K8s to restart the pod (E.g. see #321)
- The FASTEN server handles both `Exception` and `UnrecoverableError` thrown by the plug-ins.

## Motivation and context
To avoid swallowing exceptions, the FASTEN plug-ins can now handle exceptions in two ways:
- For the sake of backward compatibility, plug-ins can still use the `setPluginError` method, which is handled by the FASTEN server.
- Unhandled `Exception` in the plug-ins are now handled by the FASTEN server and the error is sent to the `*.err` Kafka topic.

## Testing
It's tested with `DummyPlugin`. It works as expected for all four cases: (1) successful processing of records (2) handled exceptions in the plug-in, i.e. `setPluginError` (3) Unhandled exceptions in the plug-in (4) Throwing `UnrecoverableError`

## Task list  
- [x] Define the `UnrecoverableError` exception.
- [x] Handle both `Exception` and `UnrecoverableError` in the FASTEN server

### DIFF
--- a/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
+++ b/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
@@ -294,12 +294,11 @@ public class FastenKafkaPlugin implements FastenServerPlugin {
                     plugin.consume(record.value());
                 }
             }
-        } catch (RuntimeException e) {
+        } catch (UnrecoverableError e) {
             // In rare circumstances, plug-ins throw UnrecoverableError to crash and therefore K8s will restart the plug-in.
-            if (e instanceof UnrecoverableError) {
-                logger.error("Forced to stop the plug-in due to ", e);
-                throw e;
-            }
+            logger.error("Forced to stop the plug-in due to ", e);
+            throw e;
+        } catch (RuntimeException e) {
             logger.error("An error occurred in " + plugin.getClass().getCanonicalName(), e);
             plugin.setPluginError(e);
         }
--- a/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
+++ b/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
@@ -19,6 +19,7 @@
 package eu.fasten.server.plugins.kafka;
 
 import com.google.common.base.Strings;
+import eu.fasten.core.exceptions.UnrecoverableError;
 import eu.fasten.core.plugins.KafkaPlugin;
 import eu.fasten.server.plugins.FastenServerPlugin;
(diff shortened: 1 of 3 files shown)


server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -264,32 +265,43 @@ public void handleConsuming() {
...
+                }
             }
-        } else { // If local storage is not enabled.
-            if (consumeTimeoutEnabled) {
-                consumeWithTimeout(record.value(), consumeTimeout, exitOnTimeout);
-            } else {
-                plugin.consume(record.value());
+        } catch (RuntimeException e) {
+            // In rare circumstances, plug-ins throw UnrecoverableError to crash and therefore K8s will restart the plug-in.
+            if (e instanceof UnrecoverableError) {
```
proksch, line 299: Please do not use this check. Instead have a separate catch clause for `UnrecoverableError` first, and then have a second catch clause for the more general `RuntimeException`. The clauses will be processed in order and only the first match is being executed.

proksch (changes requested):

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Addressed in e116cdb.

mir-am:
@proksch,
Thanks for reviewing the PR.
I've addressed your comment.
Let me know if we can merge this PR.

proksch:
Can you please rebase the changes onto develop? Also, please try to avoid merge commits as 
59d1325 in the future, this is being resolved by a rebase as well.

mir-am:
> Can you please rebase the changes onto develop? Also, please try to avoid merge commits as [59d1325](https://github.com/fasten-project/fasten/commit/59d132551660e9744eef3a28397dbfe3be4b8689) in the future, this is being resolved by a rebase as well.

Yes! I have addressed your comments and the PR is tested with `DummyPlugin` for all four possible cases as explained in the testing section of the main post.

proksch:
"This branch cannot be rebased due to conflicts".

It is not clear to me what you did, but you did not rebase. ;)

mir-am:
> "This branch cannot be rebased due to conflicts".
> 
> It is not clear to me what you did, but you did not rebase. ;)

I'm gonna resolve the conflicts and rebase onto `develop`.

PR #332, opened 2021-11-16
Remove default consume topic from MetadataDBExtension

amottier:
## Description
Remove default value for the topic to consume from in `MetadataDBExtension` class. Add default value in `MetadataDBJavaExtension` class.
Also make `MetadataDBExtension` abstract to avoid accidental direct usage.

## Motivation and context
Default value for the topic to consume from make sense only in actual plugin implementation (i.e. C, Java and Python).
It was already set for C and Python, adding it to Java.

## Testing
Build project and run unit test. No extra tests added.

### DIFF
--- a/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
+++ b/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
@@ -51,9 +51,9 @@ import java.sql.BatchUpdateException;
 import java.sql.Timestamp;
 import java.util.*;
 
-public class MetadataDBExtension implements KafkaPlugin, DBConnector {
+public abstract class MetadataDBExtension implements KafkaPlugin, DBConnector {
 
-    protected List<String> consumeTopics = new LinkedList<>(Collections.singletonList("fasten.OPAL.out"));
+    protected List<String> consumeTopics = null;
     private static DSLContext dslContext;
     protected boolean processedRecord = false;
     protected Exception pluginError = null;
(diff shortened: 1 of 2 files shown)


proksch:
I am honestly wondering, why we have these default topics in the code at all... I would make it mandatory to have them in one place only, the deployment files. It is super confusing for newcomers, if they change the topic and cannot see any effect, just because the deployment file has a higher priority.

I would propose that all plugins should require a set of named input/output topics that have to be provided in the deployment or else the FastenServer would not even start. This would a) allow to have multiple instances running on different pipelines (as in the previous approach for normal/priorities queues) and b) make config explicit and avoid redundant information.

proksch:
Should we try to remove the default values from all plugins?

amottier:
> Should we try to remove the default values from all plugins?

I think this make sense. Also I recommend to use same topic name in k8s and in docker-compose. I spot some incoherence and tried to highlight them on the [deployment diagram](https://github.com/fasten-project/fasten/wiki/Deployment).

I can work on that once everyone agree on the proposal.

proksch:
I am not sure how strict we have to be with the naming consistency... the beauty of being able to move the topics to the deployment is that the pipelines can look differently, e.g. a minimalistic debugging Docker compose setup and a "full" deployment on k8s... this might reflect in the names somehow.

To add a c) to my previous point: Changing this would also allow us to have multiple input/output topics. We are already half-way through the recent PR #330 but making these dependencies explicit, would allow us to detect (and warn about) misconfiguration on startup...

mir-am:
I generally agree with the proposal. We can create a separate PR to remove default Kafka topics from the plug-ins and also force the FASTEN server to make input topics mandatory.

mir-am (approved):
LGTM.

PR #339, opened 2021-11-25
Fixing CommitFailedException in the FASTEN server

mir-am:
## Description
This PR addresses Kafka's `CommitFailedException` by making the following changes:
- Storing The last time the normal consumer's `poll()` method is called and compare it to the value of ` getMaxConsumeTimeout()`. Using this comparison, we call the `poll()` method of the normal consumer at least one periodically depending on the value of ` getMaxConsumeTimeout()`. This should prevent `CommitFailedException` .
- To handle process signals such as `SIGTERM`, register a shutdown hook to wake up the consumer and finally close the Kafka connections gracefully.

## Motivation and context
With recent changes to the FASTEN server (see #330), two Kafka consumers, namely, normal and priority are introduced. Given that the priority consumer has precedence over the normal consumer, the poll() method of the normal consumer may not be called after `max.poll.interval.ms`, which causes Kafka's `CommitFailedException`.
Also, deleting K8s deployments sends SIGTERM signal, which was not previously handled in the FASTEN server and again it may cause Kafka's `CommitFailedException`

## Testing
Using `DummyPlugin`, I have managed to reproduce the issue and also tested the fix. 

## Task list  
- [x] Register shutdown hook to gracefully close Kafka connections
- [x] Fix `CommitFailedException` by calling the poll method of the normal consumer periodically.

### DIFF
--- a/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
+++ b/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
@@ -150,7 +150,37 @@ public class FastenKafkaPlugin implements FastenServerPlugin {
 
     @Override
     public void run() {
-        if (plugin.consumeTopic().isPresent()) {
+        subscribeToTopics();
+        
+        if (this.skipOffsets == 1) {
+            skipPartitionOffsets();
+        }
+
+        try {
+            while (!isConnectionsClosed.get()) {
+                try {
+                	if (plugin.consumeTopic().isPresent()) {
+	                	handleConsuming();
+	                } else {
+	                    doCommitSync(KafkaRecordKind.NORMAL);
+	                    handleProducing(null, System.currentTimeMillis() / 1000L, KafkaRecordKind.NORMAL);
+	                }
+	            } catch (WakeupException e) {
+	                if (isConnectionsClosed.get()) {
+		                // WakeupException is used to interrupt long-running polls, e.g., we are use it
+	                	// to interrupt polls when handling shutdown signals. The exception can be ignored.
+	                }
+	            }
+            }
+        } finally {
+            connNorm.close();
+            connPrio.close();
+            logger.info("Plugin {} stopped gracefully", plugin.name());
+        }
+    }
+
+	private void subscribeToTopics() {
+		if (plugin.consumeTopic().isPresent()) {
             normTopics = plugin.consumeTopic().get().stream().filter(s -> !s.contains("priority")).collect(Collectors.toList());
             connNorm.subscribe(normTopics);
 
@@ -159,31 +189,7 @@ public class FastenKafkaPlugin implements FastenServerPlugin {
                 connPrio.subscribe(prioTopics);
             }
         }
-        if (this.skipOffsets == 1) {
-            skipPartitionOffsets();
-        }
-
-        try {
-            while (!isConnectionsClosed.get()) {
-                if (plugin.consumeTopic().isPresent()) {
-                    handleConsuming();
-                } else {
-                    doCommitSync(KafkaRecordKind.NORMAL);
-                    handleProducing(null, System.currentTimeMillis() / 1000L, KafkaRecordKind.NORMAL);
-                }
-            }
-        } catch (WakeupException e) {
-            // Wakeup exception is caught after handling shutdown signals, i.e., deleting deployments in Kubernetes.
-            // This exception is used to wake up Kafka consumers/clients
-            if (!isConnectionsClosed.get()) {
-                throw e;
-            }
-        } finally {
-            connNorm.close();
-            connPrio.close();
-            logger.info("Plugin {} stopped", plugin.name());
-        }
-    }
+	}
 
     /**
      * Starts the plugin.
@@ -205,7 +211,7 @@ public class FastenKafkaPlugin implements FastenServerPlugin {
     public void handleConsuming() {
         boolean hasConsumedPriorityRecord = false;
 
-        // Keep the normal consumer alive when there are tons of priority records to process
+        // Refresh connection timeout of normal consumer when priority records are processed
         sendHeartBeat(connNorm);
 
         if (!prioTopics.isEmpty()) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -168,8 +173,12 @@ public void run() {
                     handleProducing(null, System.currentTimeMillis() / 1000L, KafkaRecordKind.NORMAL);
                 }
             }
-        } catch (Exception e) {
-            logger.error("Error occurred while processing call graphs", e);
+
+        } catch (WakeupException e) {
+            if (!closed.get()) throw e;
```
proksch, line 178: A comment would help here to understand in which cases the exception does not need to be re-thrown. What is the semantic of `!closed.get()`?

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -168,8 +173,12 @@ public void run() {
...
+
+//        } catch (Exception e) {
```
proksch, line 180: Commented out code should always be removed (unless it serves a purpose, like documentation). Commented code is not subject to refactorings and as such will rot. The VCS history is enough to revert back to old versions...

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -191,6 +200,18 @@ public void stop() {
         closed.set(true);
     }
 
+    /**
+     * Whether the poll() method of the normal Kafka connection should be called to avoid CommitFailedException considering
+     * the `max.poll.interval.ms` time, i.e., consumeTimeout value,
+     */
+    private boolean shouldNormPollBeCalled() {
+        if (this.lastTimeNormPollCalled != null) {
```
proksch, line 208: I dislike the idea of keeping track of the time, as this does not guarantee a general solution to the problem. Should the Kafka config be changed to a smaller timeout, we will see the same problem. 

As a general solution to this, Kafka seems to support the concept of heartbeats... I would suggest to further explore the option to just sent a heartbeat for both normal and prio queue consumers at the beginning of each loop to prevent Kafka from rebalancing.

proksch: Plus, as long as there are prio messages, we should not waste computational power on normal messages...

proksch (changes requested):
I dislike the approach for how these types of exceptions are handled. I think we should explore the option of sending heartbeats first, as that would present a more general solution to the problem.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Addressed in bdb0474.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: I've found a better solution, which is using the `pause()` and `resume()` methods. See 54bde0c. I've tested with DummyPlugin and it works nicely!

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
proksch: Please rephrase: "WakeUpException is used to wake up waiting Kafka clients. If the connection is already closed, we can ignore the exception, otherwise, we re-throw it."

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -209,7 +213,13 @@ public void handleConsuming() {
...
         }
 
-        if (!normTopics.isEmpty() && !hasConsumedPriorityRecord) {
+        if (!normTopics.isEmpty()) {
+            // Pause the normal consumer if there are priority records to process.
+            if (hasConsumedPriorityRecord) {
+                connNorm.pause(connNorm.assignment());
+            } else {
+                connNorm.resume(connNorm.assignment());
+            }
```
proksch, line 222: Couldn't we just introduce an artificial "heartbeat" method like:



    private static void sendHeartBeat(KafkaConsumer<String, String> con) {
    	// see https://stackoverflow.com/a/43722731
    	con.pause(con.assignment());
    	con.poll(Duration.ZERO);
    	con.resume(con.assignment());
    }

And call this in the beginning of the poll loop? These three calls should be cheap, but refresh the timeout counter. I would prefer this solution over a complication of the already complex logic in the rest of the method...

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -579,4 +588,15 @@ public boolean isConsumeTimeoutEnabled() {
...
     }
+
+    /**
+     * It cleans up resources after receiving the SIGTERM signal. E.g. closing Kafka connections
+     */
+    private void registerShutDownHook() {
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            this.stop();
+            this.connNorm.wakeup();
+            logger.info("Cleaned up resources before shutting down the JVM");
```
proksch, line 599: This message is incorrect... the WakeupException only breaks long-running polls, which allows the plugin to exit timely. Maybe just remove the message.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -579,4 +588,15 @@ public boolean isConsumeTimeoutEnabled() {
...
         return consumeTimeout;
     }
+
+    /**
+     * It cleans up resources after receiving the SIGTERM signal. E.g. closing Kafka connections
+     */
+    private void registerShutDownHook() {
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            this.stop();
+            this.connNorm.wakeup();
```
proksch, line 598: Add the same call for connPrio

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -168,8 +170,10 @@ public void run() {
                     handleProducing(null, System.currentTimeMillis() / 1000L, KafkaRecordKind.NORMAL);
                 }
             }
-        } catch (Exception e) {
-            logger.error("Error occurred while processing call graphs", e);
+
+        } catch (WakeupException e) {
```
proksch, line 174: As discussed offline: the WakeupExceptions should be caught in a try catch that *only* wraps the contents of the loop and, if caught, should be swallowed. Please elaborate in the empty catch that the exception just short-cuts long-running polls.  If we move the try/catch, we do not need to handle this outside anymore, when we should actually quit, `closed` is set and ends the `while`.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
proksch: (see comments below)

proksch (changes requested):

mir-am:
@proksch
Thanks for the minor refactoring.
I have tested the new changes with `DummyPlugin` and it works as expected.
We can rebase and merge this PR.

PR #340, opened 2021-11-25
Make CallableIndexServerPlugin to read GID graphs only from FS

mir-am:
## Description
Now, `CallableIndexServerPlugin` reads GID graphs only from the file system and throws `RuntimeException` if a GID graph is not found.

## Motivation and context
There is currently an inconsistency between GID graphs on FS and Kafka topics, which should be avoided.
Moreover, the metadata DB plug-in writes GID graphs only to the disk to reduce the size of its output Kafka records.

## Testing
None yet.

## Task list  
- [x] Read GID graphs from FS, otherwise throw `RuntimeException`.
- [x] Fix unit tests


analyzer/callable-index/src/test/java/eu/fasten/analyzer/callableindex/CallableIndexServerPluginTest.java
```diff
@@ -44,37 +48,35 @@ public void setUp() {
...
-                "\"edges\": [[0, 1], [1, 2]]" +
-                "}}";
-        callableIndexFastenPlugin.consume(json);
+    public void consumeTest() throws FileNotFoundException {
+        var json = new JSONObject("{\"payload\": {}}");
+
+        var getTestResource = Objects.requireNonNull(Thread.currentThread().getContextClassLoader().getResource("gid_graph_test.json"));
+        json.getJSONObject("payload").put("dir", getTestResource.getPath());
+        callableIndexFastenPlugin.consume(json.toString());
         assertNull(callableIndexFastenPlugin.getPluginError());
```
proksch, line 69: When we refactor the class, we should revise the error handling to the "new" (normal) way. Exceptions should be thrown and caught... whatever handling the `FastenServer` introduces outside of the logic here when wrapping the plugin does not matter and we should avoid this weird registration of errors as much as possible.

analyzer/callable-index/src/test/java/eu/fasten/analyzer/callableindex/CallableIndexServerPluginTest.java
```diff
@@ -44,37 +48,35 @@ public void setUp() {
...
     }
 
     @Test
     public void consumeJsonErrorTest() {
-        callableIndexFastenPlugin.consume("{\"payload\":{\"foo\":\"bar\"}}");
+        var json = new JSONObject("{\"payload\": {}}");
+
+        var getTestResource = Objects.requireNonNull(Thread.currentThread().getContextClassLoader().getResource("gid_graph_test_err.json"));
```
proksch, line 76: See comment above. This needs to be fixed throughout the class.

analyzer/callable-index/src/test/java/eu/fasten/analyzer/callableindex/CallableIndexServerPluginTest.java
```diff
@@ -44,37 +48,35 @@ public void setUp() {
...
-                "\"version\": \"0.0.1\"," +
-                "\"nodes\": [1, 2, 3]," +
-                "\"numInternalNodes\": 2," +
-                "\"edges\": [[1, 2], [2, 3]]" +
-                "}}");
-        var graph = GidGraph.getGraph(json.getJSONObject("payload"));
+        var json = new JSONObject("{\"payload\": {}}");
+
+        var getTestResource = Objects.requireNonNull(Thread.currentThread().getContextClassLoader().getResource("gid_graph_test.json"));
+        json.getJSONObject("payload").put("dir", getTestResource.getPath());
```
proksch, line 54: I am not a big fan of storing the JSON on disk versus declaring it in the class, but I acknowledge that there are benefits/drawbacks for both solutions and I do not have a strong preference... in this concrete case, however, we need to change contents (add the temp dir), so I would argue that keeping everything in one place makes it easier to understand here. I would suggest to revert back to the old declaration as string. What do you folks think?

proksch (changes requested):
I have my doubts about this change, please see the different inline comments for details..

analyzer/callable-index/src/main/java/eu/fasten/analyzer/callableindex/CallableIndexServerPlugin.java
```diff
@@ -112,17 +111,7 @@ public void consume(String record) {
...
-                        gidGraph = ExtendedGidGraph.getGraph(json);
-                    } else {
-                        gidGraph = GidGraph.getGraph(json);
-                    }
-                } catch (JSONException e) {
-                    logger.error("Could not parse GID graph", e);
-                    setPluginError(e);
-                    return;
-                }
+                throw new RuntimeException("Couldn't find the GID graph file!");
```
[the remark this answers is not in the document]
mir-am: Addressed in 137d330. Also, if the file doesn't exist on the FS, the FNF exception is caught and the exception is set.

analyzer/callable-index/src/test/java/eu/fasten/analyzer/callableindex/CallableIndexServerPluginTest.java
```diff
@@ -44,37 +48,35 @@ public void setUp() {
...
-                "\"product\": \"test\"," +
-                "\"version\": \"0.0.1\"," +
-                "\"nodes\": [1, 2, 3]," +
-                "\"numInternalNodes\": 2," +
-                "\"edges\": [[1, 2], [2, 3]]" +
-                "}}");
-        var graph = GidGraph.getGraph(json.getJSONObject("payload"));
+        var json = new JSONObject("{\"payload\": {}}");
+
+        var getTestResource = Objects.requireNonNull(Thread.currentThread().getContextClassLoader().getResource("gid_graph_test.json"));
```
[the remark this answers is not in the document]
mir-am: Done!

analyzer/callable-index/src/test/java/eu/fasten/analyzer/callableindex/CallableIndexServerPluginTest.java
mir-am: Done!

analyzer/callable-index/src/test/java/eu/fasten/analyzer/callableindex/CallableIndexServerPluginTest.java
mir-am: Addressed in 6ba5d88.

analyzer/callable-index/src/test/java/eu/fasten/analyzer/callableindex/CallableIndexServerPluginTest.java
mir-am: Since the plug-in now reads GIDs from the FS, it expects the JSON input to have a path rather than the whole graph. If you know a trick to pass the GID graph as a string for testing let me know.

proksch: Ok, let's postpone this discussion for now to the point when we focus our refactoring efforts to the CallableIndexer

proksch (approved):

Issue #338, opened 2021-11-23
Clean-up of MavenCoordinate required


proksch:
Right now, we have the class [MavenCoordinate](https://github.com/fasten-project/fasten/blob/develop/core/src/main/java/eu/fasten/core/data/opal/MavenCoordinate.java) to abstract away some Maven details... Unfortunately, the class is a big mess and urgently requires a thorough refactoring.

- It is supposed to be a data structure, but it contains a download method that downloads Maven artifact from a repository. This should be split up into two classes. The data structure and a utils class.
- The test for the class are still located in the `javacg-opal` plugin, apparently the class has been moved, but the tests were left in the old place.
- The tests of `CallGraphConstructor` are based on the downloading functionality and *download actual packages on every build*, this is not acceptable. I propose two extensions 1) downloads should be cached in a configurable folder, which should also reduce the network footprint of the FASTEN pipeline. 2) The download logic in `MavenCoordinate` should be refactored such that it can be mocked for the `CallGraphConstructorTest` to avoid a download and make the test faster.

Do we have any volunteer that wants to clean up this class?

mir-am:
I can take care of the first two tasks. We have `MavenUtilties` in `core` which downloads Maven artifacts.
I leave the third task to Mehdi as he's the main developer for the OPAL plug-in.

mir-am:
The first and second tasks were done in #344. I'd like to close the issue. We can do the third task in another issue or PR.

proksch:
Agreed, but then please open another issue or this will be forgotten. ;)

PR #344, opened 2021-11-29
Minor refactoring to the MavenCoordinate class

mir-am:
## Description
- Creates a separate class for `MavenResolver` and renames it to `MavenArtifactDownloader` to avoid confusion with the Graph `MavenResolver` in core.
- Move `MavenCoordinate` and `MavenArtifactDownloader` tests to core from the OPAL module.

## Motivation and context
As stated in #338, it is better to separate the JAR downloads logic from the `MavenCoordinate` class and also moves its tests to core.

## Testing
Unit tests for `MavenCoordinate` and the JAR downloader pass successfully.

## Task list  
- [x] Separate the JAR downloader logic from `MavenCoordinate`
- [x] Move unit tests for MavenCoordinate and `MavenArtifactDownloader` to core.


analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/data/MavenCoordinateTest.java
```diff
@@ -1,115 +0,0 @@
-/*
```
proksch, line 1: Why has this file been deleted? It is just in the wrong place... Instead of deleting it, it should be moved to FASTEN core (where the Maven Coordinate is located).

core/src/main/java/eu/fasten/core/data/opal/MavenArtifactDownloader.java
```diff
@@ -0,0 +1,121 @@
...
+            "aar", "car", "nar", "kar"};
+    private static final String[] defaultPackaging = {"zip", "aar", "tar.gz", "jar"};
+
+    /**
+     * Download a JAR file indicated by the provided Maven coordinate.
+     *
+     * @param mavenCoordinate A Maven coordinate in the for "groupId:artifactId:version"
+     * @return A temporary file on the filesystem
+     */
+    public File downloadArtifact(final MavenCoordinate mavenCoordinate, String artifactRepo)
```
proksch, line 33: I would suggest to break this method down into several smaller helper methods... the complexity of the method goes through the roof!

core/src/main/java/eu/fasten/core/data/opal/MavenArtifactDownloader.java
```diff
@@ -0,0 +1,121 @@
...
+            in.close();
+
+            return Optional.of(new File(tempFile.toAbsolutePath().toString()));
+        } catch (IOException e) {
+            if (tempFile != null) {tempFile.toFile().delete();}
+            throw new MissingArtifactException(e.getMessage(), e.getCause());
+        }
+    }
+
+    private long computeDurationInMs(long startTime) {
```
proksch, line 117: Taking the time for the download should be removed, this is unrelated for the task at hand. If this is necessary for a specific case, I would solve it through implementing through subclassing and overriding the corresponding method (the override can then instruments it with timing logic).

core/src/main/java/eu/fasten/core/data/opal/MavenArtifactDownloader.java
```diff
@@ -0,0 +1,121 @@
...
+
+            long startTime = System.nanoTime();
+
+            try {
+                if (Arrays.asList(packaging).contains(mavenCoordinate.getPackaging())) {
+                    found = true;
+                    jar = httpGetFile(mavenCoordinate
+                            .toProductUrl(repos.get(i), mavenCoordinate.getPackaging()));
+                }
+            } catch (MissingArtifactException e) {
```
proksch, line 51: As we are cycling through multiple repositories, the log message should boil down to a single info entry that explains that artifacts was not found in repo XY and that we retry the next one... if it is not found in any repo, we can throw a `MissingArtifactException` at the end.

Related to that, I noticed that several repos are being tried before Maven Central, but as it is the most common, I would argue that Maven Central should be the first hit in the list... I am not sure though where in the code the repos are hard-coded though.

proksch (changes requested):
Good first step, but more refactoring is required.

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/data/MavenCoordinateTest.java
mir-am: Sorry, forgot to push the tests for them in core. It's here a0ba202.

core/src/main/java/eu/fasten/core/data/opal/MavenArtifactDownloader.java
mir-am: Addressed in 5f4cad7.

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/CGandStitchingTest.java
```diff
@@ -340,8 +341,7 @@ private ExtendedRevisionJavaCallGraph getRCG(String coordStr)
         throws OPALException, MissingArtifactException {
         final var coord = MavenCoordinate.fromString(coordStr,"jar");
 
-        final var dep1 =
-            new MavenCoordinate.MavenResolver().downloadArtifact(coord,"jar");
+        final var dep1 = new MavenArtifactDownloader().downloadArtifact(coord,"jar");
```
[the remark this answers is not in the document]
mir-am: Done!

core/src/main/java/eu/fasten/core/data/opal/MavenArtifactDownloader.java
mir-am: I am not sure whether we can remove the timing info there. Wouter changed the format of OPAL's logs and also, IIRC included timing for his monitoring solution.

core/src/main/java/eu/fasten/core/data/opal/MavenArtifactDownloader.java
mir-am: Source repos are specified in the input records of OPAL, i.e., the `artifactRepository` field. I think it tries the Maven Central repo first most of the time as most input records have Maven Central defined as the source repo.

mir-am:
@proksch, thanks for your contribution.
Deleted tests in a8dd5e4 are NOT duplicate. I moved them from the OPAL module to the core module.

PR #346, opened 2021-11-30
Fixes to make tests succeed on Windows (10/11).

MagielBruntink:
## Description
The recent changes to OPAL included some breaking tests on Windows. Since I like Windows, I fixed them :-)


proksch (approved):
I was not aware of the TestUtils class. How handy!

MagielBruntink:
> I was not aware of the TestUtils class. How handy!

It is, but it's perhaps not in the right place.

PR #348, opened 2021-11-30
Fix issue with vulnerability-statements-processor deleting too many mappings to package-versions and callables.

MagielBruntink:
@mir-am This fix requires additional unique indices added to the DB to work. If you already added the tables and indices for the vulnerability tables, could you please update each DB (java, c, python) with the following?
```
CREATE UNIQUE INDEX CONCURRENTLY unique_vuln_x_pkg_ver ON vulnerabilities_x_package_versions USING btree (vulnerability_id, package_version_id);
ALTER TABLE vulnerabilities_x_package_versions ADD CONSTRAINT unique_vuln_x_pkg_ver UNIQUE USING INDEX unique_vuln_x_pkg_ver;

CREATE UNIQUE INDEX CONCURRENTLY unique_vuln_x_callable ON vulnerabilities_x_callables USING btree (vulnerability_id, callable_id);
ALTER TABLE vulnerabilities_x_callables ADD CONSTRAINT unique_vuln_x_callable UNIQUE USING INDEX unique_vuln_x_callable;
```


mir-am:
Thanks for the fix. I have not deployed the new vulnerability architecture yet on monster. I'll consider this when I deploy it.

Issue #349, opened 2021-11-30
JavaCGOPAL container fails to run


MagielBruntink:
## Describe the bug
After recent changes to JavaCGOPAL its container does not properly run anymore. There are errors in the log (see below) on header fields and manifests. The final log line are:
```
[2021-11-30 20:54:25,320] [INFO ] [main] [e.f.s.FastenServer] - Plugin init done: 0 KafkaPlugins, 0 DB plug-ins, 0 GraphDB plug-ins: 0 total plugins
[2021-11-30 20:54:25,325] [INFO ] [main] [e.f.s.FastenServer] - No DB connection info provided, skipping initialization.
```
which is especially weird since I don't think that JavaCGOpal connects to the DB.

## To Reproduce
1. `docker pull ghcr.io/fasten-project/fasten/fasten.server.develop.slim:latest`
2. `FASTEN_MVN_REPO=https://repo.maven.apache.org/maven2/ docker run ghcr.io/fasten-project/fasten/fasten.server.develop.slim:latest -k localhost:9092 -cg OPALPlugin_2 -pl OPAL -kt OPAL=fasten.POMAnalyzer.out -po OPAL=/tmp/revision-callgraphs`

## Logs
```
[2021-11-30 20:28:45,428] [INFO ] [main] [e.f.s.FastenServer] - Plugin init done: 0 KafkaPlugins, 0 DB plug-ins, 0 GraphDB plug-ins: 0 total plugins
[2021-11-30 20:28:45,429] [INFO ] [main] [e.f.s.FastenServer] - No DB connection info provided, skipping initialization.
[2021-11-30 20:28:48,945] [INFO ] [main] [e.f.s.FastenServer] - FASTEN server started in development mode
[2021-11-30 20:28:48,957] [INFO ] [main] [e.f.s.FastenServer] - ************************System Info***************************
[2021-11-30 20:28:49,004] [INFO ] [main] [e.f.s.FastenServer] - Max Heap size: 4294967296 Bytes
[2021-11-30 20:28:49,004] [INFO ] [main] [e.f.s.FastenServer] - **************************************************************
[2021-11-30 20:28:49,056] [INFO ] [main] [o.p.DefaultPluginStatusProvider] - Enabled plugins: []
[2021-11-30 20:28:49,059] [INFO ] [main] [o.p.DefaultPluginStatusProvider] - Disabled plugins: []
[...]

PR #350, opened 2021-12-01
Reduce plugin description to one line to fix manifest error

proksch:
Fixes the issue reported in #349.

The problem is hilarious: when the plugin description is broken down to multiple lines (which is perfectly fine in the declaring XML), the resulting `MANIFEST.MF` contains new lines that the parser does not like. In the particular case, the line was too long, so the auto-formatter added a new line.

The simple solution was to shorten the description to make sure that we do not see this issue happening again (accidentally). Maybe we should think about removing the description altogether from all plugins, what do they add?


MagielBruntink (approved):
Wow, great find! Indeed, let's get rid of those descriptions, lest they cause more harm in the future.

I built the image locally and now it runs properly. Cheers!

PR #351, opened 2021-12-02
Adding endpoints for C & Python Dependents Resolution

gdrosos:
## Description
Extended the REST-API plugin to support two new endpoints for C & Python Dependents Resolution. 
The 2 endpoints have the following format:
  ```/pypi/packages/{pkg}/{pkg_ver}/resolve/dependents``` 
  ```/debian/packages/{pkg}/{pkg_ver}/resolve/dependents```
  and follow the exact same logic with the corresponding endpoint of Maven Dependents Resolution.

## Motivation and context
Since we now can resolve the C & Python package dependents throught #324 , we want to share this utility throught the REST-API Plugin.

## Testing
Tested all the endpoints locally and they have the expected behaviour.

## Additional context
  * In order to implement the endpoints, the path to the Python & C Metadata DB is required, as well as the path of the serialized Dependency Graph of each language. Those new values are retrieved from the [application.properties](https://github.com/fasten-project/fasten/blob/python-c-dependents-rest-api/analyzer/restapi-plugin/src/main/resources/application.properties) file which is initialized through Kubernetes. Thus, in order to expose the updated REST-API, we need to update the REST-API plugin k8s configuration files with the correct new paths.
  * This pull request is based on #324, and therefore it should be merged after it.


analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/GraphResolversInitializer.java
```diff
@@ -0,0 +1,31 @@
+package eu.fasten.analyzer.restapiplugin.mvn;
+
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.client.RestTemplate;
+
+public class GraphResolversInitializer {
```
mir-am, line 6: This might be a silly question!
Why sending a request to initialize graph resolvers? These should be initialized asynchronously when the REST API server is started. Right?

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/DebianResolutionApiServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    private GraphResolver graphDebianResolver;
+
+    public DebianResolutionApiServiceImpl() {
+        try {
+            var graphDebianResolver = new GraphResolver();
+            graphDebianResolver.buildDependencyGraph(KnowledgeBaseConnector.dbCContext, KnowledgeBaseConnector.dependencyDebianGraphPath);
+            this.graphDebianResolver = graphDebianResolver;
+        } catch (Exception e) {
+            logger.error("Error constructing Debian dependency graph resolver", e);
+            System.exit(1);
```
mir-am, line 48: Why do you stop the REST API if an exception happens? Perhaps you can return a custom message to the user and show that an error happened when resolving dependents. That is, something like this `return ResponseEntity<String>(msg)`

mir-am:
Moreover, please fix the build error for `c066263`.

gdrosos:
> Moreover, please fix the build error for `c066263`.

The build error happens because this Rest-API extension uses some data structures proposed on #324, and therefore the build will be successful only after the merging of the previous PR.

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/DebianResolutionApiServiceImpl.java
gdrosos: Ι stopped the REST-API because I followed the same logic with that of [ResolutionAPIService class](https://github.com/fasten-project/fasten/blob/ba6d41e64c627b04be8e523782fa2f28a50b7554/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/ResolutionApiServiceImpl.java#L65) but if its required I can change it to the proposed return message.

gdrosos:
Thank you very much @mir-am  for the immediate review! I tried answering your comments above. The logic behind the Dependents endpoints is exactly the same with the endpoint of the  [Maven Dependents](https://github.com/fasten-project/fasten/blob/develop/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/ResolutionApiServiceImpl.java#L106). After merging this PR, we will have all the required endpoints for C & Python. (Microservices for C/Python Stitching, Microservices for Debian/PyPi Dependency Resolution, and finally with this PR 2 endpoints for Dependents Resolution).

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/GraphResolversInitializer.java
proksch: The problem is that the server normally only starts when all endpoints are initialized. Thsi means that startup takes 10min when the graph has to be built. If we make everything lazy, then the startup is instant, but the first request takes 10min. We decided to do a compromise and start lazy, but immediately place a request for the graph, so the graph building starts right away. It still takes 10min for the graph to be available after startup, but the rest of the API is already available in this time.

PR #355, opened 2021-12-03
Vulnerability restapi purl

cg122:
## Description
Added one more endpoint `POST /vulnerabilities/purls` to directly get vulnerabilities from already process vulnerability statement.

## Task

- [x] implemented new endpoint
- [x] tested with local docker-deployment instance


mir-am:
Thanks for the fix.
Unfortunately, this is pretty hard for me to review as there are 54 commits in this PR from other developers.
I would suggest first pulling the latest `develop` branch, creating a new branch, applying your changes, and finally pushing only your commits.
Maybe @MagielBruntink has a better solution for this.

MagielBruntink:
Sorry, I just merged it. Obviously in the future we need to be more careful with branching off, rebasing, etc. But this is a relatively small change by itself and it was already tested through.

MagielBruntink:
And that created a huge mess in the commit history... :-( Why did the merging re-apply the commits??

Issue #356, opened 2021-12-03
JavaCGOpal fails to produce, throws errors only


MagielBruntink:
## Describe the bug
After recent changes, the JavaCGOpal container does not create revision call graphs correctly, throwing "scala.MatchError: Seq (of class java.lang.String)".

Surprisingly, all unit tests succeed.

## To Reproduce
1. Clone https://github.com/fasten-project/fasten-docker-deployment
2. `docker-compose --profile java pull`
3. `docker-compose --profile java up -d`
4. `cat test-resources/mvn.coords.txt | docker-compose exec -T kafka kafka-console-producer.sh --broker-list kafka:19092 --topic fasten.mvn.releases --property 'parse.key=true' --property 'key.separator=|'`

## Logs
```
{
   "input": {
      "input": {
         "groupId": "log4j",
         "artifactId": "log4j",
         "version": "1.2.17"
      },
      "plugin_version": "0.1.2",
      "consumed_at": 1638546198,
      "payload": {
         "date": 1338025419000,
         "repoUrl": "http://svn.apache.org/viewvc/logging/log4j/tags/v1_2_17_rc3",
         "groupId": "log4j",
         "version": "1.2.17",
         "parentCoordinate": "",
         "artifactRepository": "https://repo.maven.apache.org/maven2/",
         "forge": "mvn",
         "sourcesUrl": "https://repo.maven.apache.org/maven2/log4j/log4j/1.2.17/log4j-1.2.17-sources.jar",
         "artifactId": "log4j",
         "dependencyData": {
            "dependencyManagement": {
               "dependencies": []
            },
            "dependencies": [
               {
                  "versionConstraints": [
                     {
                        "isUpperHardRequirement": false,
                        "isLowerHardRequirement": false,
                        "upperBound": "1.4.3",
                        "lowerBound": "1.4.3"
                     }
                  ],
                  "groupId": "javax.mail",
                  "scope": "",
                  "classifier": "",
                  "artifactId": "mail",
                  "exclusions": [],
                  "optional": true,
                  "type": ""
               },
               {
                  "versionConstraints": [
                     {
                        "isUpperHardRequirement": false,
                        "isLowerHardRequirement": false,
                        "upperBound": "5.0-2",
                        "lowerBound": "5.0-2"
                     }
                  ],
                  "groupId": "org.apache.openejb",
                  "scope": "provided",
                  "classifier": "",
                  "artifactId": "javaee-api",
                  "exclusions": [],
                  "optional": false,
[...]

PR #358, opened 2021-12-05
Delete extra reference.conf in the server project

proksch:
The fasten-server project wrongly contained an OPAL configuration file. Removing this file should solve issue #356.


MagielBruntink (changes requested):
This change seems to cause another issue, unfortunately:
```[2021-12-06 08:58:16,642] [ERROR] [pool-1-thread-1] [e.f.a.j.OPALPlugin$OPAL] - [CG-GENERATION] [FAILED] [1455] [log4j:log4j:1.2.17] [OPALException] com.typesafe.config.ConfigException$Missing: withValue(opalj.br.reader.ClassFileReader.Invokedynamic.rewrite): No configuration setting found for key 'org.opalj.tac'```

Kafka message on the error topic:
```
{
   "input": {
      "input": {
         "groupId": "log4j",
         "artifactId": "log4j",
         "version": "1.2.17"
      },
      "plugin_version": "0.1.2",
      "consumed_at": 1638781094,
      "payload": {
         "date": 1338025419000,
         "repoUrl": "http://svn.apache.org/viewvc/logging/log4j/tags/v1_2_17_rc3",
         "groupId": "log4j",
         "version": "1.2.17",
         "parentCoordinate": "",
         "artifactRepository": "https://repo.maven.apache.org/maven2/",
         "forge": "mvn",
         "sourcesUrl": "https://repo.maven.apache.org/maven2/log4j/log4j/1.2.17/log4j-1.2.17-sources.jar",
         "artifactId": "log4j",
         "dependencyData": {
            "dependencyManagement": {
               "dependencies": []
            },
            "dependencies": [
               {
                  "versionConstraints": [
                     {
                        "isUpperHardRequirement": false,
                        "isLowerHardRequirement": false,
                        "upperBound": "1.4.3",
                        "lowerBound": "1.4.3"
                     }
                  ],
                  "groupId": "javax.mail",
                  "scope": "",
                  "classifier": "",
                  "artifactId": "mail",
                  "exclusions": [],
                  "optional": true,
                  "type": ""
               },
               {
                  "versionConstraints": [
                     {
                        "isUpperHardRequirement": false,
                        "isLowerHardRequirement": false,
                        "upperBound": "5.0-2",
                        "lowerBound": "5.0-2"
                     }
                  ],
                  "groupId": "org.apache.openejb",
                  "scope": "provided",
                  "classifier": "",
                  "artifactId": "javaee-api",
                  "exclusions": [],
                  "optional": false,
                  "type": "jar"
               },
               {
                  "versionConstraints": [
                     {
                        "isUpperHardRequirement": false,
                        "isLowerHardRequirement": false,
                        "upperBound": "2.0.8",
                        "lowerBound": "2.0.8"
                     }
                  ],
                  "groupId": "oro",
                  "scope": "test",
                  "classifier": "",
                  "artifactId": "oro",
                  "exclusions": [],
                  "optional": false,
                  "type": ""
               },
               {
                  "versionConstraints": [
                     {
                        "isUpperHardRequirement": false,
                        "isLowerHardRequirement": false,
                        "upperBound": "3.8.2",
                        "lowerBound": "3.8.2"
                     }
                  ],
                  "groupId": "junit",
                  "scope": "test",
                  "classifier": "",
                  "artifactId": "junit",
                  "exclusions": [],
                  "optional": false,
                  "type": ""
               },
               {
                  "versionConstraints": [
                     {
                        "isUpperHardRequirement": false,
                        "isLowerHardRequirement": false,
                        "upperBound": "1.0",
                        "lowerBound": "1.0"
                     }
                  ],
                  "groupId": "org.apache.geronimo.specs",
                  "scope": "",
                  "classifier": "",
                  "artifactId": "geronimo-jms_1.1_spec",
                  "exclusions": [],
                  "optional": true,
                  "type": ""
               }
            ]
         },
         "projectName": "Apache Log4j",
         "commitTag": "",
         "packagingType": "bundle"
      },
      "host": "6a8712963c5c",
      "created_at": 1638781095,
      "plugin_name": "POMAnalyzer"
   },
   "plugin_version": "0.1.2",
   "err": {
      "msg": "com.typesafe.config.ConfigException$Missing: withValue(opalj.br.reader.ClassFileReader.Invokedynamic.rewrite): No configuration setting found for key 'org.opalj.tac'",
      "stacktrace": [
         "eu.fasten.analyzer.javacgopal.data.OPALCallGraphConstructor.construct(OPALCallGraphConstructor.java:70)",
         "eu.fasten.analyzer.javacgopal.data.OPALCallGraphConstructor.construct(OPALCallGraphConstructor.java:49)",
         "eu.fasten.analyzer.javacgopal.data.PartialCallGraphConstructor.createExtendedRevisionJavaCallGraph(PartialCallGraphConstructor.java:120)",
         "eu.fasten.analyzer.javacgopal.OPALPlugin$OPAL.consume(OPALPlugin.java:88)",
         "eu.fasten.server.plugins.kafka.FastenKafkaPlugin.lambda$consumeWithTimeout$7(FastenKafkaPlugin.java:556)",
         "java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)",
         "java.base/java.util.concurrent.FutureTask.run(Unknown Source)",
         "java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)",
         "java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)",
         "java.base/java.lang.Thread.run(Unknown Source)"
      ],
      "error": "OPALException"
   },
   "consumed_at": 1638781095,
   "host": "7f9384276b17",
   "created_at": 1638781096,
   "plugin_name": "OPAL"
}
```

**Comment by mir-am** [comment]
Since Sebastian is on vacation, I will try to fix the issue today.

**Comment by mir-am** [comment]
Even when `reference.conf` is deleted, the file still exists in the JAR. Apparently, OPAL generates a default one while building the JAR.

**Comment by mir-am** [comment]
Okay, I've managed to reproduce the issue using the `Main` class of the OPAL plugin. 
Here are the commands (in `analyzer/javacg-opal`):
```
mvn clean install -DskipTests -P main
cd docker/plugins && java -jar javacg-opal-0.0.1-SNAPSHOT-with-dependencies.jar -a "log4j:log4j:1.2.17" -g -i COORD
```
Also, I've excluded `reference.conf` from the JAR in POM. Yet, the issue persists.
I am wondering why the unit tests for OPAL pass on both local machines and CI servers.

**Comment by proksch** [comment]
According to the OPAL guys, It is not required to include a `reference.conf` in the .jar file and the tests work without one... if there is one generated during the build, this is done by logic that we added.

**Comment by proksch** [comment]
Generally speaking, I do not think that we should treat OPAL differently to any other dependency that we include in the project... I do not like the idea to have custom exclusions in the pom file, generate special files for it, or rely on some magic that happens in the build.

As it seems we have various integrations throughout the project that accommodate for OPAL. They result in these weird issues now... I would argue that, instead of adopting all the special handlings, instead, we should remove all of them. As the tests illustrate, it is possible to use OPAL as a pure library without special treatment.

**Comment by mir-am** [comment]
> Generally speaking, I do not think that we should treat OPAL differently to any other dependency that we include in the project... I do not like the idea to have custom exclusions in the pom file, generate special files for it, or rely on some magic that happens in the build.
> 
> As it seems we have various integrations throughout the project that accommodate for OPAL. They result in these weird issues now... I would argue that, instead of adopting all the special handlings, instead, we should remove all of them. As the tests illustrate, it is possible to use OPAL as a pure library without special treatment.

I understand your point and it is fair.
The project is currently quite big, i.e., it has many dependencies and modules.
Those exclusions were added to solve the PF4j exceptions at run-time, i.e., class loading issues, etc.

**Comment by proksch** [comment]
To be honest, I still do not understand the need for PF4J. We could implement `FastenServer` as a library and get rid of PF4J... it just seems to add one level of complexity and makes debugging the plugins much harder, while I still do not understand which value it actually brings to the table. We do not need a plugin system, we always only load exactly one plugin.

**Comment by mir-am** [comment]
In the beginning, we chose PF4J as it allows loading JARs dynamically and suits our plug-in-based design.
I am open to new ideas and solutions. If you have a solution that allows us to get rid of PF4J without many changes and hours of debugging, let us know.

**Comment by mir-am** [comment]
Okay, getting back to the issue, here are my findings:
1) While deleting `reference.conf`, it still appears in the final JAR of the OPAL plug-in. It gives the following OPAL exception:
```
Caused by: com.typesafe.config.ConfigException$Missing: reference.conf @ jar:file:/Users/amir/projects/fasten/analyzer/javacg-opal/docker/plugins/javacg-opal-0.0.1-SNAPSHOT-with-dependencies.jar!/reference.conf: 1: No configuration setting found for key 'org.opalj.tac'
```
2) The `reference.conf` file can be removed from the JAR by excluding it in POM. However, it gives the same exception while it doesn't reference the file in the JAR:
```
Caused by: com.typesafe.config.ConfigException$Missing: withValue(opalj.br.reader.ClassFileReader.Invokedynamic.rewrite): No configuration setting found for key 'org.opalj.tac'
```

**Comment by proksch** [comment]
> 1. While deleting `reference.conf`, it still appears in the final JAR of the OPAL plug-in. It gives the following OPAL exception:
> 2. The `reference.conf` file can be removed from the JAR by excluding it in POM. However, it gives the same exception while it doesn't reference the file in the JAR:

As mentioned before, I am wondering WHY it appears in the .jar file. As a library you cannot just decide to add stuff to the resulting package, this is something that WE apparently do at some point. When it is there, it will be read though...

**Comment by proksch** [comment]
> In the beginning, we chose PF4J as it allows loading JARs dynamically and suits our plug-in-based design. I am open to new ideas and solutions. If you have a solution that allows us to get rid of PF4J without many changes and hours of debugging, let us know.

My current understanding is that PF4J just adds unnecessary complexity. We can keep picocli for argument parsing and use `FastenServer` as a framework. This means that we would generate one concrete .jar per plugin. This would even simplify the input parameters, as we do not need to specify anymore which topics belong to which plugin, because it is only one plugin that gets started.

But I agree, let's not tangle this discussion with the issue at hand and skip the discussion for now. I will raise this point next week in another issue once I am back.

**Comment by mir-am** [comment]
> > 1. While deleting `reference.conf`, it still appears in the final JAR of the OPAL plug-in. It gives the following OPAL exception:
> > 2. The `reference.conf` file can be removed from the JAR by excluding it in POM. However, it gives the same exception while it doesn't reference the file in the JAR:
> 
> As mentioned before, I am wondering WHY it appears in the .jar file. As a library you cannot just decide to add stuff to the resulting package, this is something that WE apparently do at some point. When it is there, it will be read though...

In the first case, we delete the config file. I believe that it is generated by OPAL again. Because the generated file is not even JSON-like config.

**Comment by mir-am** [comment]
Besides, by reading the tests log, it appears that something is clearly wrong. For example, we have so many warnings like `There is an incomplete call site that OPAL did not take care of.` or `couldn't find the stmt` by `PartialCallGraph`.
Also, how is it possible to merge CGs with ZERO edges?
```
[2021-12-06 14:34:44,053] [INFO ] [main] [e.f.c.m.CGMerger] - Merging graph with 13572 nodes and 0 edges
[2021-12-06 14:34:45,961] [INFO ] [main] [e.f.c.m.CGMerger] - Merged call graphs in 1.908 seconds, num node: 12786, num edges: 67660
```

**Comment by proksch** [comment]
> ... I believe that it is generated by OPAL again...

How would this be possible without adding a specific action to our pom?

> Because the generated file is not even JSON-like config.

So, what format does it use?

**Comment by mir-am** [comment]
> > ... I believe that it is generated by OPAL again...
> 
> How would this be possible without adding a specific action to our pom?
> 
> > Because the generated file is not even JSON-like config.
> 
> So, what format does it use?

I don't know the answer to your first question!

The format was strange to me! You can just build the JAR of the OPAL plug-in and see `reference.conf` in the root of the JAR file. Whatever the format is, it allows having comments, unlike JSON.

**Comment by MagielBruntink** [comment]
I think I have a fix inbound, at least for this issue. Please do not revert yet.

**Comment by MagielBruntink** [comment]
I have some good and bad news.

The good news:
The original issue can be fixed by changing a value in the config file: "Seq" is not a valid value, but "FIFO" would be. I committed that change to this branch. See also the related OPAL code here:
https://github.com/opalj/opal/blob/945a9a92dbf18de9ad78f73306af9d6dc27489d0/OPAL/si/src/main/scala/org/opalj/fpcf/par/PKECPropertyStore.scala#L1006

Mostly bad news:
 - the file reference.conf needs to be present in javacg-opal/src/main/resources in order for it to apply to the unit tests
 - that file needs to be present *also* in server/src/main/resources.conf in order for it to be used in the server plugin... :-(
 - with the orignal issue fixed, more configuration issues pop-up, probably because the 4.0.0 has different config requirements, eg.: `No configuration setting found for key 'org.opalj.br.analyses.cg.callBySignatureResolution`

```
"err": {
      "msg": "com.typesafe.config.ConfigException$Missing: reference.conf @ jar:file:/server-0.0.1-SNAPSHOT-with-dependencies.jar!/reference.conf: 13: No configuration setting found for key 'org.opalj.br.analyses.cg.callBySignatureResolution'",
      "stacktrace": [
         "eu.fasten.analyzer.javacgopal.data.OPALCallGraphConstructor.construct(OPALCallGraphConstructor.java:70)",
         "eu.fasten.analyzer.javacgopal.data.OPALCallGraphConstructor.construct(OPALCallGraphConstructor.java:49)",
         "eu.fasten.analyzer.javacgopal.data.PartialCallGraphConstructor.createExtendedRevisionJavaCallGraph(PartialCallGraphConstructor.java:120)",
         "eu.fasten.analyzer.javacgopal.OPALPlugin$OPAL.consume(OPALPlugin.java:88)",
         "eu.fasten.server.plugins.kafka.FastenKafkaPlugin.lambda$consumeWithTimeout$7(FastenKafkaPlugin.java:556)",
         "java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)",
         "java.base/java.util.concurrent.FutureTask.run(Unknown Source)",
         "java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)",
         "java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)",
         "java.base/java.lang.Thread.run(Unknown Source)"
      ],
      "error": "OPALException"
   }
```

I think the way out for this is two-fold:
1. resolve the situation with needing to have the config file in two places
2. work with the OPAL people to figure out what configuration changes are needed for 4.0.0.

**Comment by MagielBruntink** [comment]
@ashkboos Could you also have a look at this?

**Comment by MagielBruntink** [comment]
Well, it turns out adding "callBySignatureResolution": "false" to reference.conf made it work. However, I really feel we should do a thorough review of this massive configuration. And also to have it in one place!

**Comment by ashkboos** [comment]
> Well, it turns out adding "callBySignatureResolution": "false" to reference.conf made it work. However, I really feel we should do a thorough review of this massive configuration. And also to have it in one place!

I dynamically loaded the whole default config in the debug mode. You can do this by adding a breakpoint in the OPALCallGraphConstructor construct method and looking into the project variable and seeing the config field of it. It is 1905 lines, there are a couple of new configuration options that we are missing now (such as enforceJava11CallSemantics). I added them and I'm building the project now, if it worked I will push it.

**Comment by MagielBruntink** [comment]
I tested this with a local deployment and graphs are indeed being generated again :-)

I am seeing both nodes and edges in the callgraphs, but in the logs there are quite a few warnings of the sort: 
```
[e.f.a.j.d.PartialCallGraph] - There is an incomplete call site that OPAL did not take care of. source:DefinedMethod(declaringClassType=org.apache.pdfbox.pdfwriter.COSWriter,definedMethod=org.apache.pdfbox.pdfwriter.COSWriter{ private void doWriteSignature() })PCs: [333]
```

@ashkboos Do you think the OPAL people would be available to review our configuration in light of our goals?

ashkboos:
> [e.f.a.j.d.PartialCallGraph] - There is an incomplete call site that OPAL did not take care of. source:DefinedMethod(declaringClassType=org.apache.pdfbox.pdfwriter.COSWriter,definedMethod=org.apache.pdfbox.pdfwriter.COSWriter{ private void doWriteSignature() })PCs: [333]

This is because of the call-sites that OPAL sometimes skips due to not having enough information in the partial analysis. We can disable this warning if it's annoying. I expect that it is happening less frequently in the new version due to their bug fixes.

> @ashkboos Do you think the OPAL people would be available to review our configuration in light of our goals?

I think they would be available. We had a call and showed them our code 2 weeks ago. I plan to run some experiments with the new version and contact them again so that they can do more improvements in their next release. Their current plan is to release version 5 somewhere in the spring. If you have any specific questions or pointers for discussion please let me know.

proksch:
I specifically asked about the config last time we talked and they said it is optional, hence my efforts to get rid of a monstrosity of a config that we have no idea about... Defaults plus our three explicit settings are good enough. The working tests are a proof of concept: The config is optional and only used when found.

I apologize for leaving this issue unresolved before my vacation and agree with @mir-am , please just revert the change, I will take care of this next week.

proksch:
I have debugged the issue further and the culprit seems to be our usage of the Maven shade plugin, which repackages all jar contents of dependencies into our .jar file. For whatever reason, this seems to break the OPAL plugin, which comes with some pre-packaged `reference.conf` file.

I have created a minimal working example, in which I change the packaging: Instead of re-packaging all dependencies, I put them into a `lib` folder:

    myjar.jar
    lib/abstract-interpretation-framework_2.12-4.0.0.jar
    lib/annotations-13.0.jar
    lib/...

The `myjar.jar` will still contain a `MANIFEST.MF` that refers to the main class, plus it contains valid class path entries that refer to the lib folder:

    Manifest-Version: 1.0
    Created-By: Apache Maven ${maven.version}
    Build-Jdk: ${java.version}
    Main-Class: myjar.Main
    Class-Path: lib/abstract-interpretation-framework_2.12-4.0.0.jar lib/annotations-13.0.jar ...

As a result, the jar is still executable (as long as it is placed next to said lib folder). 

A minor disadvantage of this solution is that the overall file size grows from 90MB to about 140MB, as we cannot selectively exclude jar contents anymore, however, it also has clear advantages:

1) Packaging becomes *MUCH* faster (even for my tiny toy project I can achieve a 3.5x speedup in build time; 3.6s instead of 12.1s)
2) We get rid of all former "defines overlapping classes" warnings during the build
3) OUR jar only contains contents that WE are responsible for. Likewise, we do not mess with the original jar files of our dependencies and leave their internal structure intact
4) Our Maven config becomes much simpler, as we can avoid the complex shade plugin
5) Most importantly, it works! :D

If this proposed solution is acceptable for everybody, I would look into adopting the various `Dockerfile`s and our configs of Maven and the GitHub build. To finally fix this issue for good.

proksch:
> ...
> A minor disadvantage of this solution is that the overall file size grows from 90MB to about 140MB, as we cannot selectively exclude jar contents anymore, however, it also has clear advantages:
> ...

Just to clarify. `myjar.jar` will be reduced to a couple of KB big... but the `lib` folder in total requires quite a bit of space. I think we can further optimize that by, for example, excluding test-related dependencies, but we simply require some heavy hitters like the rocksdb jar or the fastutils:

    $ ls -S
    -rw-r--r--   1 seb staff  33M Dec 14 12:47 rocksdbjni-6.15.2.jar
    -rw-r--r--   1 seb staff  23M Dec 14 12:47 fastutil-8.5.2.jar
    -rw-r--r--   1 seb staff 9.8M Dec 14 12:47 fastutil-8.5.4-javadoc.jar
    -rw-r--r--   1 seb staff 7.5M Dec 14 12:47 fastutil-extra-8.5.4.jar
    -rw-r--r--   1 seb staff 5.4M Dec 14 12:47 fastutil-core-8.5.4.jar
    -rw-r--r--   1 seb staff 5.2M Dec 14 12:47 scala-library-2.12.13.jar
    -rw-r--r--   1 seb staff 4.7M Dec 14 12:47 bytecode-representation_2.12-4.0.0.jar
    ...

mir-am:
Thanks for the investigation and finding a fix. Does the solution work with the FASTEN server and PF4J?

amottier:
I was never a big fan of using Maven shade plugin and I very much prefer this solution that allow us to keep an easy to understand packaging configuration and a clean packaged software.

proksch:
This, I still have to figure out when I am trying to adopt the changes from the minimal example to the actual project. I do not expect any problems, but I cannot judge what PF4J is doing, this is probably the most tricky part. I am not a fan of PF4J though, so if it makes any issues, getting rid of it is also on the table ;)

proksch:
Ok, after playing with PF4J for the whole day, it seems that it can only work with fat jars (or similar packaging structures like .zip files that combine classes and dependencies). This is very disappointing and essentially makes the suggested solution impossible.

As a quick fix for the current problem, we will simply provide an up-to-date `reference.conf` file that we generated from the default config. This is not a clean solution, but it will get everything up and running again.... the PF4J refactoring is then a larger endeavor that will take some more time. I will open a separate PR for it, once I have made some progress on that front.

mir-am:
@MagielBruntink
I've added the default config files of OPAL to the project again like before. The OPAL call graph generator is working now both locally and on monster.
Perhaps you can also test the fix. If it's okay, please feel free to merge the fix.

proksch:
@mir-am Right now the branch has conflicts. Please rebase and solve these conflicts first.

mir-am:
> @mir-am Right now the branch has conflicts. Please rebase and solve these conflicts first.

@proksch 
Actually, I have tried to resolve the conflicts to rebase onto develop but It was unsuccessful! It keeps complaining about the same conflicts over and over again even after resolving them and `git rebase --continue`.
I very much appreciate it if you could try rebasing. Maybe I'm doing something wrong.

proksch:
Ok, I will look into it...

MagielBruntink:
Tested in the local pipeline as well, it's back to working order.

PR #359, opened 2021-12-06
Fixes tests for callable-index on Windows 10/11.

MagielBruntink:
## Description
Use getTestResource to make tests succeed also on Windows 10/11. Also lifted the mocking of RocksDB to the class level, so some tests also succeed.


mir-am (approved):
LGTM.

PR #360, opened 2021-12-06
Adjust API response to prettify rendering of callables in web

romatallinn:
## Description
In the API response of list of callables for the module, instead of raw Fasten URI, include the parsed object of it.

## Motivation and context
In order for the web app to be more dynamic and clear in how it renders callables for the users, it needs Fasten URI to be parsed into pieces (method name, arguments and return types, etc), so the web can display them accordingly and functionally rich.

## Additional context
Related: [fasten-project/fasten-web!66](https://github.com/fasten-project/fasten-web/pull/66)

### DIFF
--- a/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
+++ b/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
@@ -19,7 +19,10 @@
 package eu.fasten.core.data.metadatadb;
 
 import com.github.t9t.jooq.json.JsonbDSL;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import eu.fasten.core.data.Constants;
+import eu.fasten.core.data.FastenJavaURI;
 import eu.fasten.core.data.metadatadb.codegen.Keys;
 import eu.fasten.core.data.metadatadb.codegen.enums.Access;
 import eu.fasten.core.data.metadatadb.codegen.enums.CallableType;
@@ -1051,37 +1054,18 @@ public class MetadataDao {
                 .limit(limit)
                 .fetch();
 
-        // Returning the result
         logger.debug("Total rows: " + queryResult.size());
-        var res = queryResult.formatJSON(new JSONFormat().format(true).header(false).recordFormat(JSONFormat.RecordFormat.OBJECT).quoteNested(false));
 
-
-        //// Insert user-friendly formatted method signature
-
-        // Parse result json string back into object
-        JSONArray json;
-        try {
-            json = new JSONArray(res);
-        } catch (JSONException err) {
-            logger.error("Error JSON Parser: " + err.toString());
-            return null;
+        List<FastenJavaURI> items = new ArrayList<>();
+        for (var partialUri : queryResult.getValues(Callables.CALLABLES.FASTEN_URI)) {
+            var fullUri = FastenUriUtils.generateFullFastenUri("mvn", packageName, packageVersion, partialUri);
+            var uriObj = new FastenJavaURI(fullUri);
+            items.add(uriObj);
         }
 
-        // Go through each callable, parse fasten uri, insert signature.
-        for (Object j : json) {
-            JSONObject jObj = (JSONObject) j;
-            var uri = jObj.getString("fasten_uri");
-
-            try {
-                var uriObject = FastenUriUtils.parsePartialFastenUri(uri);
-                jObj.put("method_name", uriObject.get(2));
-                jObj.put("method_args", uriObject.get(3));
-            } catch (IllegalArgumentException err) {
-                logger.warn("Error FASTEN URI Parser: " + err.toString());
-            }
-        }
-
-        return json.toString();
+        GsonBuilder builder = new GsonBuilder();
+        Gson gson = builder.create();
+        return gson.toJson(items);
     }
 
     public String getPackageBinaryModules(String packageName, String packageVersion, int offset, int limit) throws PackageVersionNotFoundException {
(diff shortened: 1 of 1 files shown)


core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
```diff
@@ -1067,20 +1067,22 @@ public String getModuleCallables(String packageName,
...
-
             try {
-                var uriObject = FastenUriUtils.parsePartialFastenUri(uri);
-                jObj.put("method_name", uriObject.get(2));
-                jObj.put("method_args", uriObject.get(3));
+                JSONObject jObj = (JSONObject) j;
+                var partialUri = jObj.getString("fasten_uri");
+                var fullUri = FastenUriUtils.generateFullFastenUri("mvn", packageName, packageVersion, partialUri);
+                var uriObj = new FastenJavaURI(fullUri);
+                var js =  gson.toJson(uriObj);
```
romatallinn, line 1080: Although we discussed it multiple times that the implementation of API must not depend on Java, it currently does. In this example, a higher level of refactoring needs to be done for it to be independent (e.g., pass forge along from API, etc). I suggest leaving it like this for now. I will come back to refactoring API as a whole later.

romatallinn, line 1080: Although we discussed it multiple times that the implementation of API must not depend on Java, it currently does. In this example, a higher level of refactoring needs to be done for it to be independent (e.g., pass forge along from API, etc). I suggest leaving it like this for now. I will come back to refactoring API as a whole later.

romatallinn:
@ashkboos @mir-am do we want to inform stakeholders about this small change if we merge it, right?

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
```diff
@@ -1067,20 +1067,22 @@ public String getModuleCallables(String packageName,
...
+        Gson gson = builder.create();
+
         // Go through each callable, parse fasten uri, insert signature.
         for (Object j : json) {
-            JSONObject jObj = (JSONObject) j;
-            var uri = jObj.getString("fasten_uri");
-
             try {
-                var uriObject = FastenUriUtils.parsePartialFastenUri(uri);
-                jObj.put("method_name", uriObject.get(2));
```
mir-am, line 1077: What would be the side effect of removing these two fields, i.e., `method_name`? Maybe the partners are using them.

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
mir-am: A stupid question! What would be the benefit of `FullFASTENUri` for a pretty rendering of callables?

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
romatallinn: The field was included by me for web use. Not sure if any of the partners are using it. Tbh, I doubt it. The value can still be found in the JSON, but just not by `method_name`, but in `fasten_uri/functionOrAttributeName`.

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
romatallinn: > 

Although it does include a few properties that are not really needed for my use, they are still nice to have around.

romatallinn:
@proksch do you have any thoughts? Can we merge it? In simple words, the change is seen here (before, after):

Screenshot 2021-12-06 at 17 20 38
Screenshot 2021-12-06 at 17 20 17

ashkboos (approved):

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
```diff
@@ -1067,20 +1067,22 @@ public String getModuleCallables(String packageName,
...
-                jObj.put("method_args", uriObject.get(3));
+                JSONObject jObj = (JSONObject) j;
+                var partialUri = jObj.getString("fasten_uri");
+                var fullUri = FastenUriUtils.generateFullFastenUri("mvn", packageName, packageVersion, partialUri);
+                var uriObj = new FastenJavaURI(fullUri);
+                var js =  gson.toJson(uriObj);
+                jObj.put("fasten_uri", new JSONObject(js));
```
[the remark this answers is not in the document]
romatallinn: Haha I didn't even notice it, but believe me or not, it works as expected. I assume because objects are reference types, so jObj references entities inside json.

But I do agree it's f'ed up in terms of readability and maintainability. I thought of any other way of writing this logic, but everything comes up a bit more messy or with more code. Would a comment explaining that it changes json through referencing be enough? Or you do want me to re-write this code completely?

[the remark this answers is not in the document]
ashkboos: Well, just define a jsonarray called result and add jObj to it in the last line of the for loop and return result instead. The current implementation is very misleading. Also adding comments is not a clean solution, code should be self-explanatory. Instead of adding 3 lines of comments adding these 3 lines of code is better IMO.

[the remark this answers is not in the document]
romatallinn: @ashkboos changed. Please, have a look.

ashkboos (approved):

PR #365, opened 2021-12-14
Fix vulnerable (transitive) dependencies

MagielBruntink:
## Description
FASTEN dependencies have several known vulnerabilities in them. This PR fixes those. 

See the attached CSV file for the list:
[fasten-all-vulnerabilities.csv](https://github.com/fasten-project/fasten/files/7711021/fasten-all-vulnerabilities.csv)

## Motivation and context
Vulnerabilties are bad.

## Testing
Ran unit tests successfully.

## Task list  
- [ ] Validate whether all vulnerabilities are resolved. With FASTEN itself :-)
- [x] Succesfull CI run.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-compress</artifactId>
-                <version>1.20</version>
+                <version>1.21</version>
             </dependency>
             <dependency>
                 <groupId>org.jsoup</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -88,33 +88,53 @@
         <module>server</module>
     </modules>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-compress</artifactId>
+                <version>1.20</version>
+            </dependency>
+            <dependency>
+                <groupId>org.jsoup</groupId>
+                <artifactId>jsoup</artifactId>
+                <version>1.14.2</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.logging.log4j</groupId>
+                <artifactId>log4j-api</artifactId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -88,33 +88,53 @@
         <module>server</module>
     </modules>
 
+    <dependencyManagement>
+        <dependencies>
```
proksch, line 92: Why were these additional dependencies added?

mir-am: And why here?
We add these common dependencies (like `log4j`) to the server module so that all the plug-ins can use them.

mir-am (approved):
Thanks for updating the dependencies. Overall, LGTM. I have just one minor comment.

pom.xml
MagielBruntink: The depedencyManagement section controls versions of (transitive) dependencies used. These few are needed here to upgrade vulnerable dependencies of shrinkwrap (commons-compress, jsoup, log4j-api). If the dependencyManagement section is added to the POM of fasten/core it doesn't control the versions of modules under fasten/analyzer, annoyingly. So it needs to go here, in the parent POM.

mir-am: I see!

MagielBruntink:
> Thanks for updating the dependencies. Overall, LGTM. I have just one minor comment.

I don't see log4j in the server/pom.xml. It's down somewhere else as a transitive dep.

pom.xml
proksch: Oh, I was too fast in skimming the changes and missed this detail...

MagielBruntink:
We could consider a more elaborate updating of dependencies in this PR. I saw several that could use a new version, but without a vulnerability present. Also there is a lot of duplication in the inclusions of deps all over the place. Do we want to make an effort here to clean that up?

proksch:
I think this would be a great idea. I would just postpone this until we have finalized the updates to the build infrastructure that we are currently discussing in #358 .

PR #369, opened 2021-12-15
Simplify Extended Revision Call Graph Classes Structure

gdrosos:
## Description
This Pull Request aims to simplify the Extended Revision Call Graph classes design & architecture on the following ways:

### Move Class Hierarchy Handling from ERCG Superclass to Subclasses

The [ExtendedRevisionCallGraph class](https://github.com/fasten-project/fasten/blob/develop/core/src/main/java/eu/fasten/core/data/ExtendedRevisionCallGraph.java)  contained unnecessary hierarchy information (which is language specific). In order to make it more simple, we removed those (mostly generic) attributes and implemented them in each ercg corresponding subclass (C & Python & Java), which is language specific.
### Simplify Extended Builders
Modified the extended builders -which are employed to create the Ercgs- to support the new ERCG structure.

### Remove some unused methods

Some redundant methods were removed.

## Motivation and context

The [ExtendedRevisionCallGraph class](https://github.com/fasten-project/fasten/blob/develop/core/src/main/java/eu/fasten/core/data/ExtendedRevisionCallGraph.java) was more complex than it should be (e.g. contained unnessecarry generics declarations, heavily depended on subclasses, not easily readable), whereas its main purpose is to offer a language independent structure that will be used to hold some attributes common among all languages.

## Testing
Modified Tests to support the new ERCG structure. Also tested the C MetadataPlugin and it works fine with the new Structure, I will also plan on testing the Python Metadata Plugin


gdrosos:
This is a draft Pull Request and it may be enriched with more code refactorings. For example, in both C & Python none of the plugins create the ERCGs using  the ExtendedBuilder classes ([ExtendedBuilderC](https://github.com/fasten-project/fasten/blob/develop/core/src/main/java/eu/fasten/core/data/ExtendedBuilderC.java) & [ExtendedBuilderPython](https://github.com/fasten-project/fasten/blob/develop/core/src/main/java/eu/fasten/core/data/ExtendedBuilderPython.java) ) and they are not used in any part of the pipeline. Thus, we can remove them. What is your opinion on this change? @proksch  @ashkboos

proksch:
I am all in for removing everything that does not get stored in the database. Moreover, I vote for not storing language specific context information in the database that is only relevant for the stitching. I think we should simplify the stitching by store such information on the file system in data structures that are specialized for each language.

ashkboos (approved):

ashkboos:
I would also suggest that we also get rid of RevisionCallGraph class. And rename the ERCGs to RCGs. We did this once in the past but I think other branches got merged somehow and the change was discarded!

cg122:
> Removed nodeCount attribute from ERCG classes since it is not stored in the database or used in any part of the pipeline.

I think we used this method in one of our FASTEN integration tool internally. @MagielBruntink

gdrosos:
> > Removed nodeCount attribute from ERCG classes since it is not stored in the database or used in any part of the pipeline.
> 
> I think we used this method in one of our FASTEN integration tool internally. @MagielBruntink

Thanks for pointing this out. We reverted this change accordingly.

gdrosos:
Hello,
 following Thursday's Dev Call, I removed the nodeCount attribute since it is not stored in the Metadata Database. Instead, I created a helper method called getNodeCount() which can be employed to calculate the total number of nodes. Now the ERCGs contain only attributes stored in the database.
 @cg122, @MagielBruntink feel free to employ this helper method in your internal tools to calculate the total number of nodes of each ERCG.

Also, this PR is ready for review, so @proksch feel free to take a look at it at your convenience.

MagielBruntink:
Thank you for the heads up @gdrosos this seems like a good solution to me.

PR #371, opened 2021-12-16
Disabled repeated merger efficiency tests due to build pipeline slowdown

MagielBruntink:
Disabled repeated merger efficiency tests due to build pipeline slowdown. Also focus the remaining test on efficiency and not node/edge count.


mir-am (approved):
LGTM!

Issue #367, opened 2021-12-14
Processing transitive dependencies in the FASTEN pipeline


mir-am:
Following our discussion in #352, I would like to open a new issue to discuss a solution for processing transitive dependencies in the pipeline. The solution extends POMAnalyzer as follows:

1- First, check if the given pkg. version is already inserted into DB.
2- If not, let POMAnalyzer process the given pkg. version and get its direct dependencies.
3- When adding direct deps. to the DB,
  3.1- Check if the DB already contains the package version OR the package is already ingested.
  3.2- Otherwise, proceed to step 4.
4- Direct deps. will recursively be added to `fasten.mvn.priority.ingest` and will be processed further down the pipeline, i.e. OPAL, etc.

**Remark**: The alternative to the first step would be to insert a new field to messages of `fasten.mvn.priority.ingest`, i.e., a flag to whether process a package version or not.

I am open to new ideas and suggestions. Let me know your thoughts, @MagielBruntink, @proksch.

Issue #352, opened 2021-12-02
Use lazy ingestion in POMAnalyzer for newly discovered dependencies


mir-am:
Currently, POMAnalyzer inserts newly discovered dependencies for Maven packages as a package into the metadata DB.
See:
https://github.com/fasten-project/fasten/blob/03c4ef0bf9924af2f4d25183a61a9cfc96a29286/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java#L258

We can insert this as a package version and also use a REST API call to lazy ingest it so that we will have CGs and GIDs.

MagielBruntink:
@mir-am I gave the processing of dependencies some further thought. If we have the POM Analyzer request ingestion dependencies through the REST API, pretty soon we will have a massive number of package-version to processes. Because it will request dependencies of dependencies, etc. and the REST API will put them all on the priority topics. I think it will be better to do this on the normal topic, so directly from the POM Analyzer.

mir-am:
Good point!

proksch:
I realized that the current behavior is likely the explanation for the problem raised in #347 and I am afraid that I can only think about a different solution...

The problem is simple. If the dependencies of an artifact will only be processed *after* the artifact itself, then we have problems further down in the pipeline: when we try to stitch the merged callgraph for said artifact to create the callable index, we can only consider those dependencies that are already in the database, i.e., those that have already been required by other artifacts earlier. Unknown dependencies will simply not be part of the CG merging. This is a problem!

I can only think about one single solution to this: when ingesting an artifact, we first need to check for all (direct) dependencies whether they have been ingested already and add those that have not been before processing the requested artifact. This processing needs to be recursive, as the same problem exists for direct dependencies as well (we first need to ingest their dependencies and so on). Simply resolving the dependency set does not work, as order is not considered there.

This also means that the dependencies need to be added on the same "lane" (norm/prio) than the requested artifact, as we must enforce processing order. This will become faster, the more complete the database is, but in the beginning this will likely introduce a bit of lag for lazy ingested artifacts.

[...]

PR #372, opened 2021-12-16
Extending POMAnalyzer to process full transitive dependencies

mir-am:
## Description
This PR addresses #367 and makes the following changes:
- An optional working set for the FASTEN plug-ins. This allows to plug-in to process new records based on a previously processed record.
- Adapt the FASTEN server to process records in the working set of plug-ins. After processing a normal/priority record, the server checks whether the working set is available and not empty, if yes, it processes the working set before committing the previously processed Kafka record.
- Use the working set in POMAnalyzer to add the full transitive dep. set of package versions and process them further down the pipeline, i.e. CG generation, etc.

## Motivation and context
As discussed in #352, full (transitive) dep. set of package versions is not being processed in the FASTEN pipeline, although the direct dependencies are inserted into the metadata DB. Therefore, it's necessary to identify the full (transitive) dep. set of package versions and then process them all. Otherwise, we won't be able to use `CGMerger` to merge a package version with its full dep. set.

## Testing
I've tested this solution using the FASTEN Docker compose locally and the pkg. version `org.apache.pdfbox:pdfbox:2.0.24`. Here is the results of testing before/after the fix:

|  	| Before the fix 	| After the fix 	|
|:---:	|:---:	|:---:	|
| # records in the output topic 	| 1 	| 102 	|
| # pkg. versions in the DB 	| 2 	| 67 	|
| # Dependencies in the DB 	| 10 	| 104 	|

## Task list  
- [x] Add working set to the FASTEN plug-in interface
- [x] Adapt the FASTEN server to process the working set 
- [x] Use the working set in POMAnalyzer to process the full dep. set of packager versions

## Additional context
Using the working set for POMAnalyzer, there are two blatant downsides:
1- The working set is in memory. Hence it'll be gone after a crash or restart.
2- A package version will be committed after processing its optional working set. Therefore, if the plug-in crashes or gets restarted, then the package version plus its full dep. set will be re-processed in the pipeline. This causes duplication in the downstream Kafka topics and hence wastes computational resources.


mir-am:
@proksch, @MagielBruntink
I have implemented a solution to process the full transitive dep. set of package versions in order as we discussed in #352.
It is also tested and the results are shown here above in the testing section.
I highly appreciate it if you could review the PR.

MagielBruntink:
Before I dive into the code: is there a mechanism to record what package-versions have been ingested already and are somewhere in the pipeline?

proksch:
My understanding right now is that the POMAnlyzer will create concrete entries in Postgres DB, so you should be able to look up whether the processing has started by looking up whether a concrete package/version exists in the corresponding tables... this does not give you an indication though whether the pipeline has finished for that package/version.

The additional ingestion table is just there to prevent countless duplications through REST requests.

proksch:
Maybe a stupid question/idea that came to me after reading the comment... why are we actually passing all the details in the messages in the Kafka topics? This nesting bloats the message size and makes it hard to track progress. Couldn't we just simplify the messages in all topics to just the `package_version` and record the detailed message in a key/value DB? Every topic gets its own table, the keys are the package_version, the value is the plain message, and we avoid nesting.

This would make it easily possible to check in linear time for every topic whether a certain key (package_version) already exists (= was already processed). An additional benefit is that the various analyzers only have to deserialize exactly the information that they require and not more, which might help to reduce the overhead introduced through Kafka.

MagielBruntink:
> Maybe a stupid question/idea that came to me after reading the comment... why are we actually passing all the details in the messages in the Kafka topics? This nesting bloats the message size and makes it hard to track progress. Couldn't we just simplify the messages in all topics to just the `package_version` and record the detailed message in a key/value DB? Every topic gets its own table, the keys are the package_version, the value is the plain message, and we avoid nesting.
> 
> This would make it easily possible to check in linear time for every topic whether a certain key (package_version) already exists (= was already processed). An additional benefit is that the various analyzers only have to deserialize exactly the information that they require and not more, which might help to reduce the overhead introduced through Kafka.

We discussed removing the dependencies information before, which is the biggest chunk. I think we agreed to do it, but I'm unsure whether we did.

There are some critical bits of information passed on that downstream plugins use (the following is from the top of my mind and may be incomplete):
 - forge
 - groupId, artifactId, version, product (obviously)
 - path to sources ("sourcesUrl")
 - path to call graph ("dir")

proksch:
These additional info could be easily looked up in said tables, couldn't they? I am not sayign that we shoudl just remove all information, I am saying that instead of passing it around in the message, put it into a DB for lookup. So, once you have the package_version in your hand, you can use this to request any previously computed information in linear time (as long as you know the name of the analyzer that created that).

For example, to get information about the sources location, you would look up `package-1.2.3` in the `RepositoryCloner` table and find the url there.

MagielBruntink:
> These additional info could be easily looked up in said tables, couldn't they? I am not sayign that we shoudl just remove all information, I am saying that instead of passing it around in the message, put it into a DB for lookup. So, once you have the package_version in your hand, you can use this to request any previously computed information in linear time (as long as you know the name of the analyzer that created that).
> 
> For example, to get information about the sources location, you would look up `mvn!foobar:mypackage:1.2.3` in the `RepositoryCloner` table and find the url there.

I’m not in favor of changing that now TBH. There are plugins such as the RAPID quality analyzers that at the moment do not connect to the DB at all. That was also the design of the FASTEN pipeline, to use Kafka messages for operational data, and put metadata, graphs etc in the DB. If we remove the dependencies and dependencyManagement field, which nobody downstream uses I believe, message size is very small already.

mir-am:
> 



> Maybe a stupid question/idea that came to me after reading the comment... why are we actually passing all the details in the messages in the Kafka topics? This nesting bloats the message size and makes it hard to track progress. Couldn't we just simplify the messages in all topics to just the `package_version` and record the detailed message in a key/value DB? Every topic gets its own table, the keys are the package_version, the value is the plain message, and we avoid nesting.
> 
> This would make it easily possible to check in linear time for every topic whether a certain key (package_version) already exists (= was already processed). An additional benefit is that the various analyzers only have to deserialize exactly the information that they require and not more, which might help to reduce the overhead introduced through Kafka.

We can use RocksDB for this purpose actually, i.e., key/value storage. However, I prefer to use Kafka for data flow between plugins. I'd say using another DB complicates the pipeline further.

proksch:
Ok. Let's keep it as it is and let's think about how do address the short-comings of the current solution with a less intrusive change. For now, I would say we should just accept the shortcomings, review this PR, and open an issue for an optimization that we can work on in the future, when we have more time... :>

MagielBruntink:
What I'm afraid of will happen is a massive amount of re-processing of common dependencies, eg. google guava. So for me the question remains whether POM analyzer checks in the DB whether the requested package-version itself, or some of its transitive dependencies in its working set have already been processed? POM Analyzer could easily check, and I think easily skip redundant package-versions.

proksch:
I agree that we should have a way to prevent recomputation, but I was assuming that something like that is in place. The downsides that Amir mentions refer to recomputation of the complete last depset only for crashes...

I will have time on Monday to look into the PR and understand what's there.

MagielBruntink:
It seems like lines 276-277 make the existence checks before adding dependencies to the working set:
https://github.com/fasten-project/fasten/blob/e0e9cff061a82b18c1281ecc58d927c53684dece/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java#L276
https://github.com/fasten-project/fasten/blob/e0e9cff061a82b18c1281ecc58d927c53684dece/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java#L277

So, if package-versions have been added to the DB before, or have been requested for ingestion by the REST API, they will not be added to the output topic of POM Analyzer, correct @mir-am ?

mir-am:
> What I'm afraid of will happen is a massive amount of re-processing of common dependencies, eg. google guava. So for me the question remains whether POM analyzer checks in the DB whether the requested package-version itself, or some of its transitive dependencies in its working set have already been processed? POM Analyzer could easily check, and I think easily skip redundant package-versions.

In fact, POMAnalyzer checks whether a pkg. version exists in the DB before adding it to its working set. See the checks here:
https://github.com/fasten-project/fasten/blob/9c9ea712d33e866e7c6ca539d3f87a7df269a641/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java#L275

mir-am:
> So, if package-versions have been added to the DB before, or have been requested for ingestion by the REST API, they will not be added to the output topic of POM Analyzer, correct @mir-am ?
Ok, let me clarify:
(1) If a package-versions exists in the DB, it'll not be processed by POMAnalyzer AGAIN.
(2) If a package-versions is requested for ingestion (via REST API), it'll be processed by POMAnalyzer eventually but the package version will not be added to its working set.

Note that POMAnalyzer process package versions either from its input Kafka topics OR its in-memory working set.

MagielBruntink:
> Ok, let me clarify: (1) If a package-version exists in the DB, it'll not be processed by POMAnalyzer AGAIN. (2) If a package-version is requested for ingestion (via REST API), it'll be processed by POMAnalyzer eventually but the package version will not be added to its working set.
> 
> Note that POMAnalyzer process package versions either from its input Kafka topics OR its in-memory working set.

Thanks @mir-am 
I think in that case the check if a package_version exists in the ingested_artifacts table is then redundant, and even a bit complicating. If a package_version was already asked by the REST API for ingestion, it will either:
 1. arrive at the POM Analyzer later, in which case the package_version was already added to the DB by the POM Analyzer, and it will be skipped at that point, or,
 2. be ahead in the pipeline, but then the package_version will exist also already in the DB, so it would be skipped anyway.

By having the POM Analyzer skip the package_versions in the ingested_artifacts table, you will effectively delay their processing, which is not what you want because you want to have a complete dependency set for merging.

MagielBruntink:
Could we please address this empty "expected" exception catch?
https://github.com/fasten-project/fasten/blob/e0e9cff061a82b18c1281ecc58d927c53684dece/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java#L207

It should at least log something meaningful.

mir-am:
> Could we please address this empty "expected" exception catch?
> 
> https://github.com/fasten-project/fasten/blob/e0e9cff061a82b18c1281ecc58d927c53684dece/analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java#L207
> 
> It should at least log something meaningful.

Yes, of course! An empty catch is a bad practice and I will address it by adding a log message.

analyzer/pom-analyzer/src/main/java/eu/fasten/analyzer/pomanalyzer/POMAnalyzerPlugin.java
```diff
@@ -262,6 +271,29 @@ public long saveToDatabase(String product, String version, String repoUrl, Strin
...
+        private void addDependenciesToWorkingSet(MetadataDao metadataDao) {
+        	
+        	for (var dep : dependencyData.dependencies) {
+        		var depProduct = dep.groupId + Constants.mvnCoordinateSeparator + dep.artifactId;
+        		
+	            var depVersions = dep.getVersion().split(Pattern.quote(","));
+	            for (String v : depVersions) {
+	                if (MavenUtilities.mavenArtifactExists(dep.groupId, dep.artifactId, v, null) &&
+	                        (metadataDao.getPackageVersion(depProduct, v) == null ||
+	                                !metadataDao.isArtifactIngested(depProduct, v))) {
```
MagielBruntink, line 283: This check should be removed - either it will unnecessarily delay dependencies from being processed, or it is redundant if the package-version has already been added to the DB (because then the other clause will be true).

proksch:
We have already discussed the PR in an internal pair programming session. The current version is not yet sufficient and requires an extension to the FastenKafkaPlugin, as it is impossible so far that one Input message causes multiple output messages. We also still need to integrate shrinkwrap as discussed before...

We started with the refactoring, but did not manage to finish it yet. I hope that we will have more news tomorrow.

proksch:
I took the liberty of refactoring the pom-analyzer as I found that the current implementation gave several opportunities for improvement. Looking at the amount of changed code, calling it a rewrite would probably be more accurate at this point.

Without claim of completeness, I was working/focussing on the following main changes:

- I have extended the `FastenServer` to allow a client to produce multiple messages as a result of a single input message. @mir-am's solution proposed to use a working set at the level of the `FastenServer`, but I did not found this solution very elegant, because it was based on strong assumptions regarding the clients, instead, I would like to keep this out of the server and leave control closer to the plugins. I am still not convinced that my refactoring is the ultimate solution (e.g., output messages can only be sent in batches and not when they are ready). However, my primary goal was to take a first step towards building a cleaner, more general solution, without affecting any of the other plugins. I think that the `FastenServer` should go through another close inspection after the holidays.
- The PomAnalyzer had a very convoluted implementation that has handled all peculiarities of the Maven POM model all by itself, including resolving dependencies and properties. I am now using the two libraries `maven-model` and `maven-model-builder` that are provided by Maven itself to first build an *effective* model of the POM, which already solves all issues with inheritance and properties, and which then also provides a much more convenient way of interacting with the files through a semantic model, so we do not have to parse XML ourselves anymore. To put things into perspective, the core POM parsing took ~800LoC before and can now be implemented in less than 150LoC.
- I was very underwhelmed by the current state of testing in the pom-analyzer as many interesting cases were simply not covered. I spent a considerable effort on improving the testability of the `pom-analyzer`. For that, I broke the two main monolithic classes down into smaller pieces with clear responsibilities. I have then provided a solid number of tests for these individual classes.

To finalize this refactoring/rewrite, I still have two finalize (and test) two more classes that are related to the database interaction (`DatabaseUtil`) and interactions with remote artifact repositories (`MavenRepositoryUtil`), but this will have to wait until after Christmas. I will update this PR as I go along.

proksch:
I am done with the implementation and things work great locally, but I am stuck with my testing efforts when I try to integrate it with the server... PF4J does not like the (admittedly hacky) way in which I instrument one of the Shrinkwrap classes to capture the resolved artifact repositories. This results in class loading issues, which represent a serious road-block right now.

I spent a considerable time debugging the Shinkwrap code and I found no easy way to get to this information out.. normally nobody is interested in the source of a package as long as it resolves, so Shrinkwrap drops the source information at some point along the way. My solution for preservation is to re-define one of their utility classes in our project and patch it such that it stores this information when it passes through. Normally this class is simply loaded first and serves as a drop-in replacement for the original class. Yes, this is hacky, but it works flawlessly until class-loading magic is performed, which is where PF4J enters the scene. Unfortunately, PF4J uses a different class loader for plugins+dependencies, which breaks this mechanic.

I have two possible solutions in mind:

- Clone the repo of this one particular Shrinkwrap package, introduce my patch directly there, package it and host the package somewhere, exclude the original dependency in our pom.xml and instead include my replacement. This way, the same class loader will be used to load all plugin-related classes
- I could try to integrate Aspect4J to instrument the classes on load (instead of patching them locally), but I am not sure whether this would be compatible with PF4J as A4J also hooks into the class loading mechanism, afaik.

Overall, this is much more complicated than it should be. It also is the second time in a short while already that I am running into major issues caused by PF4J. I start to be really annoyed by the choice of this framework, most importantly, because I still do not see its value for our project.

MagielBruntink:
@proksch I have made fixes for the path/url handling. I can push them here if you want.

MagielBruntink:
I'm observing something strange. On the Windows build server, and my local machine, the POM file for one of the resolved coordinates in the `inheritedDependency` test does not contain the artifactRepository. It seems that is different on Linux, since the test there succeeds...

 I attached the POM file I am seeing:
[javax.inject-1.pom.gz](https://github.com/fasten-project/fasten/files/7786069/javax.inject-1.pom.gz)

MagielBruntink:
What could be a cause: `javax.inject` is a (transitive) dependency of FASTEN itself, through shrinkwrap. What will not work on Windows is deleting that POM file from the ~/.m2 repository while it is also open.

MagielBruntink:
The code in `resolveDependenciesFromPom` tries to delete a jar file of a dependency in order to have it re-downloaded. What will not work (on Windows) is deleting that jar file while it is also open for reading. In this case, javax.inject is a transitive dependency of FASTEN itself (through shrinkwrap) hence it cannot be deleted while FASTEN it running...

proksch:
Then I am afraid that this logic won't work under Windows. The source repository is not available for local packages, so I have to delete it to force a new download. We could work around that by collecting *all* repositories from the POMs and checking all of them for the existence of the particular package, but a) this is expensive and b) it seemed wise to reuse the functionality of Shrinkwrap before coming up with a home-brew solution... I'd suggest let's keep this logic for now and open an issue once merged. The server runs Linux, so at least in production this should be a non-issue and we can fix it for Windows later.

The updates to `ResolutionResult` look good to me, I would just like to change the handling of the `MalformedURLException`... I am not a big fan of Java's checked exceptions. I will push the update tomorrow

Some positive news: I have found a solution for the class loading issues. As envisioned before, I have cloned the Shrinkwrap repository and managed to publish a patched version of the dependency to a separated Maven repository. I have published the patched .jar both through [GitHub Packages](https://github.com/proksch/shrinkwrap-resolver/packages/1169980) and the [GitLab Package Registry](https://gitlab.com/proksch/shrinkwrap-resolver/-/packages/4381430), but both times I only realized afterwards that the registries require authentication, even for public packages, so *you* or the build server cannot download it. I will address this tomorrow and just publish the package in a repository that allows *actually public* releases. I still have to find a good provider... I am thinking about JFrog Artifactory, but I am open for suggestions.

Once this is done, I still need to double check whether the Kafka messages are still compatible with the downstream plugins and I *could* add some more tests (the actual plugin and the DB util), but apart from that we should then be able to finally release the new PomAnalyzer.

PS: I am rebasing my feature branches to get updates from master to avoid merge commits, so I need to force-push them. Therefore, I normally work alone on my branches. I appreciate contributions, please just make sure that you delete your local copy of the branch afterwards or I cannot avoid breaking your local checkout of that branch next time I rebase.

proksch:
One minor addition: in production, all dependencies of FASTEN are bundled, so we should not load anything from `.m2`. This issue should only occur during development.

MagielBruntink:
> Then I am afraid that this logic won't work under Windows. The source repository is not available for local packages, so I have to delete it to force a new download. We could work around that by collecting _all_ repositories from the POMs and checking all of them for the existence of the particular package, but a) this is expensive and b) it seemed wise to reuse the functionality of Shrinkwrap before coming up with a home-brew solution... I'd suggest let's keep this logic for now and open an issue once merged. The server runs Linux, so at least in production this should be a non-issue and we can fix it for Windows later.

The package in question, javax-inject, is not a local package however, so I wonder a bit how this issue occurs in the first place. It seems some packages do not always include the required info in their POM. Why would a re-download of the same thing fix that?

> Some positive news: I have found a solution for the class loading issues. As envisioned before, I have cloned the Shrinkwrap repository and managed to publish a patched version of the dependency to a separated Maven repository. I have published the patched .jar both through [GitHub Packages](https://github.com/proksch/shrinkwrap-resolver/packages/1169980) and the [GitLab Package Registry](https://gitlab.com/proksch/shrinkwrap-resolver/-/packages/4381430), but both times I only realized afterwards that the registries require authentication, even for public packages, so _you_ or the build server cannot download it. I will address this tomorrow and just publish the package in a repository that allows _actually public_ releases. I still have to find a good provider... I am thinking about JFrog Artifactory, but I am open for suggestions.

Why not use the Github packages, same as for the Docker images? Our developers and the build servers already authenticate with Github in order to pull them. It would be really messy if we start relying on another repo somewhere.

MagielBruntink:
> One minor addition: in production, all dependencies of FASTEN are bundled, so we should not load anything from `.m2`. This issue should only occur during development.

A cleaner solution would be to have the POM Analyzer work in a temporary directory where it keeps its local `.m2/repository`. So there will never be issues development-time with the resolution process deleting and re-generating jars.

proksch:
The links above point to my releases in said registry, but in order to use it with Maven, you need a custom entry in your local .m2/settings.xml to work. So far, this was not necessary to setup and I find it incredibly complicated for a Maven repository... I don't understand why GitHub/GitLab decided that this is a good idea, the Intention is likely just vendor lock-in.

Apparently, it is also possible to host a Maven repository in a regular GiHub repo (not through GH Packages), which then does not require authentication. I will try that first before choosing an external repository.

Regarding the custom .m2 folder for the analyzer: I thought about this as well, but the Fasten plugin will be the only process running in the Docker container, there cannot be a conflict.

analyzer/pom-analyzer/src/main/java/org/jboss/shrinkwrap/resolver/impl/maven/MavenResolvedArtifactImpl.java
```diff
@@ -0,0 +1,272 @@
...
+	}
+    // </new>
+    
+    @Override
+    public <RETURNTYPE> RETURNTYPE as(Class<RETURNTYPE> returnType) {
+        if (returnType == null) {
+            throw new IllegalArgumentException("Type must be specified.");
+        }
+
+        final FormatProcessor<? super MavenResolvedArtifact, RETURNTYPE> processor = FormatProcessors.find(
```
MagielBruntink, line 130: My IDE flags a compile error at this line:
"no instance(s) of type variable(s) RESOLVEDTYPE exist so that Object conforms to ResolvedArtifact<RESOLVEDTYPE>"

A cast is needed to resolve it:
```
final FormatProcessor<? super MavenResolvedArtifact, RETURNTYPE> processor = (FormatProcessor<? super MavenResolvedArtifact, RETURNTYPE>) FormatProcessors.find(
    MavenResolvedArtifact.class, returnType);
```

MagielBruntink:
> Regarding the custom .m2 folder for the analyzer: I thought about this as well, but the Fasten plugin will be the only process running in the Docker container, there cannot be a conflict.

Yes, in the Docker container that is true. However, I would strongly prefer if the development-time POM analyzer and its tests do not delete jars from my default M2 repository, and then resolve them potentially differently  using Shrinkwrap. That is just asking for trouble down the line. It's a bit of luck that it even works on MacOS and Linux, which allow you to delete files that are loaded into memory.

proksch:
So are you questioning the use of Shrinkwrap? The location is not a config option of Shrinkwrap, it just uses the default location of Maven, which you can change locally in your setting.xml.... 

The file that you point out is the one shipped by Shrinkwrap, which I patched. It will be deleted from the repo in my next commit.

MagielBruntink:
> So are you questioning the use of Shrinkwrap? The location is not a config option of Shrinkwrap, it just uses the default location of Maven, which you can change locally in your setting.xml....
> 
> The file that you point out is the one shipped by Shrinkwrap, which I patched. It will be deleted from the repo in my next commit.

No, the question is whether existing jars should be deleted from a developer’s default maven repo. Look, I understand the need to have something out there quickly, but potentially breaking development environments is too high a risk.

Changing the default settings.xml will not help, because it affects all usage of maven as well.

A clean temporary directory for Shrinkwrap to work for every resolution in is the best solution. It would rule out interference of existing data, as well.

As documented here, Shrinkwrap can use system properties to control its repo location:
https://gist.github.com/kpiwko/5697999#system-properties

That will require the method that gets the repo location to query for the maven.local.repo property, which it currently doesn’t. It now is limited to settings only.

proksch:
I am sorry, but I disagree with your assessment of the situation. The contents of the `.m2/repository` are deterministic. If something is missing, then Maven/Shrinkwrap/«tool of your choice» will (re-)download it and you are good to go. Even if you get different resolution results like you hinted at before, this is no problem... you just have two different versions locally and Maven is using one version, Shrinkwrap the other one. The only thing that indeed *can* break is the artifact repository resolution on Windows, as the file cannot be deleted and therefore, no information is available... but the only thing affected is the `ResolverTest`, nothing else. It is a pity that the solution does not work in Windows, but Windows has never been a focus OS for us, so this is not a deal breaker for me. I do not see a problem in the current approach, at all.

I acknowledge your concerns though and I will do a quick check tonight, whether using the system variable is an easy fix, but if it is not, I am not willing to spend more of my vacation time on (for me) cosmetic changes. If you feel uncomfortable executing the tests, I could live with the compromise of adding a `@Disabled` annotation to `ResolverTest`, so it won't be executed on developer machines unless they explicitly enable it... In any way, I will finish the implementation over the next days, but if you think the current approach is a blocker, we need to postpone the release and discuss this problem in our next dev call in January.

And just to point out the obvious, the code is not ready for review. I just kept committing on the branch that was used by Amir initially, so it keeps popping up in the PR.

proksch:
Brief update: Deploying in a GitHub repository (and not through GitHub Packages) seems to work like a charm. I have forked the Shrinkwrap project in `fasten-project/shrinkwrap-resolver-patched` and [use a branch to host the patched .jar in a custom registry](https://github.com/fasten-project/shrinkwrap-resolver-patched/tree/repository) (`repository`). This makes it possible to access the artifacts without authentication. The code is in the same repository in branch `3.1.4-patched`.

proksch:
Ok, turns out that the "Fixed file path and URL handling." actually broke the URL generation, as it a) swallowed subfolders in the repository and b) the substring call is sensitive to the existence of ending slashes in the local .m2 folder (length +- 1), which is system/config dependent. I have re-introduced a check to unify this across systems and this seems to work again.

Now that this is (really) fixed, I could re-ingest the test coordinates in Docker compose and, unfortunately, I still get the same class-loading error with PF4J that I got before. We can no longer blame a hacky solution on my side, now the patch has been properly released and we are talking about plain Maven dependencies... The problem is easy to replicate, dockerize the Fasten Server locally and feed it with one of the test coordinates. In the PomAnalyzer you will see an exception from JBoss:

    Caused by: java.lang.IllegalStateException: Service org.jboss.shrinkwrap.resolver.impl.maven.MavenResolverSystemImpl was loaded by different classloader (org.pf4j.PluginClassLoader) then service interface org.jboss.shrinkwrap.resolver.api.maven.MavenResolverSystem (jdk.internal.loader.ClassLoaders$AppClassLoader), unable to cast classes
        at org.jboss.shrinkwrap.resolver.spi.loader.SpiServiceLoader.load(SpiServiceLoader.java:165)

I have provided a [minimum working example that illustrates that the patched .jar works](https://github.com/proksch/shrinkwrap-minimal-example), so I am a bit out of ideas what to do apart from kicking PF4J out of the window. I would appreciate some input from someone who knows PF4J better than I do... (@mir-am ? @MagielBruntink ?)

proksch:
The windows build is driving me nuts. xD I have a hard time expressing the Linux system inconsistencies in a way that is also working in windows. The right place to look at would be the `ResolutionResultTest.canGeneratePomUrlWithoutDoubleSlashes`. If this is fixed, all other failing tests will work as well.

proksch:
In a final attempt, I have merged the Shrinkwrap dependencies that exist in core and in pom-analyzer to make sure that this is not the reason for the class loading issues, but to no extent, the result is the same.

It seems that, as the server depends on core, some classes are loaded by the server already, and other classes are then loaded from within the plugin, and this is what apparently does cause the hick-ups.

I am officially out of ideas. If nobody can come up with a good proposal, I only see two ways: 1) Replace Shrinkwrap (which would REALLY hurt) or 2) De-couple PomAnalyzer from FastenServer and run it standalone.

mir-am:
> Ok, turns out that the "Fixed file path and URL handling." actually broke the URL generation, as it a) swallowed subfolders in the repository and b) the substring call is sensitive to the existence of ending slashes in the local .m2 folder (length +- 1), which is system/config dependent. I have re-introduced a check to unify this across systems and this seems to work again.
> 
> Now that this is (really) fixed, I could re-ingest the test coordinates in Docker compose and, unfortunately, I still get the same class-loading error with PF4J that I got before. We can no longer blame a hacky solution on my side, now the patch has been properly released and we are talking about plain Maven dependencies... The problem is easy to replicate, dockerize the Fasten Server locally and feed it with one of the test coordinates. In the PomAnalyzer you will see an exception from JBoss:
> 
> ```
> Caused by: java.lang.IllegalStateException: Service org.jboss.shrinkwrap.resolver.impl.maven.MavenResolverSystemImpl was loaded by different classloader (org.pf4j.PluginClassLoader) then service interface org.jboss.shrinkwrap.resolver.api.maven.MavenResolverSystem (jdk.internal.loader.ClassLoaders$AppClassLoader), unable to cast classes
>     at org.jboss.shrinkwrap.resolver.spi.loader.SpiServiceLoader.load(SpiServiceLoader.java:165)
> ```
> 
> I have provided a [minimum working example that illustrates that the patched .jar works](https://github.com/proksch/shrinkwrap-minimal-example), so I am a bit out of ideas what to do apart from kicking PF4J out of the window. I would appreciate some input from someone who knows PF4J better than I do... (@mir-am ? @MagielBruntink ?)

I'll take a look at the issue as soon as I am back from holiday.

proksch:
I tried the idea of splitting the `core` project into a `server-core` and a `core`. It is still not working, but the exception changed.

    org.jboss.shrinkwrap.resolver.api.Invokable$InvocationException: Unable to load class org.jboss.shrinkwrap.resolver.spi.loader.SpiServiceLoader with class loader jdk.internal.loader.ClassLoaders$AppClassLoader@14b0e127
    	at org.jboss.shrinkwrap.resolver.api.Invokable.loadClass(Invokable.java:66)
    	at org.jboss.shrinkwrap.resolver.api.Invokable.<init>(Invokable.java:41)
    	at org.jboss.shrinkwrap.resolver.api.ResolverSystemFactory.createFromUserView(ResolverSystemFactory.java:76)
    	at org.jboss.shrinkwrap.resolver.api.ResolverSystemFactory.createFromUserView(ResolverSystemFactory.java:54)
    	at org.jboss.shrinkwrap.resolver.api.Resolvers.use(Resolvers.java:75)
    	at org.jboss.shrinkwrap.resolver.api.maven.Maven.resolver(Maven.java:36)
    	at eu.fasten.analyzer.pomanalyzer.utils.Resolver.resolvePom(Resolver.java:90)
    	at eu.fasten.analyzer.pomanalyzer.utils.Resolver.lambda$resolveDependenciesFromPom$2(Resolver.java:55)
    	at java.base/java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Unknown Source)
    	at java.base/java.util.stream.IntPipeline$Head.forEach(Unknown Source)
    	at eu.fasten.analyzer.pomanalyzer.utils.Resolver.resolveDependenciesFromPom(Resolver.java:54)
    	at eu.fasten.analyzer.pomanalyzer.POMAnalyzerPlugin$POMAnalyzer.process(POMAnalyzerPlugin.java:141)
    	at eu.fasten.analyzer.pomanalyzer.POMAnalyzerPlugin$POMAnalyzer.consume(POMAnalyzerPlugin.java:100)
    	at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.processRecord(FastenKafkaPlugin.java:302)
    	at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.handleConsuming(FastenKafkaPlugin.java:239)
    	at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.run(FastenKafkaPlugin.java:163)
    	at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.start(FastenKafkaPlugin.java:198)
    	at java.base/java.util.ArrayList.forEach(Unknown Source)
    	at eu.fasten.server.FastenServer.run(FastenServer.java:202)
    	at picocli.CommandLine.executeUserObject(CommandLine.java:1919)
    	at picocli.CommandLine.access$1100(CommandLine.java:145)
    	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
    	at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
    	at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
    	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2159)
    	at picocli.CommandLine.execute(CommandLine.java:2058)
    	at eu.fasten.server.FastenServer.main(FastenServer.java:435)
    Caused by: java.lang.ClassNotFoundException: org.jboss.shrinkwrap.resolver.spi.loader.SpiServiceLoader
    	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(Unknown Source)
    	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(Unknown Source)
    	at java.base/java.lang.ClassLoader.loadClass(Unknown Source)
    	at org.jboss.shrinkwrap.resolver.api.Invokable.loadClass(Invokable.java:64)
    	... 26 common frames omitted

The offending class is definitely in the class path of `pom-analyzer` (dependency `shrinkwrap-spi`). It seems that we have a serious class loading issue here, but what I do not understand is why it only occurs with Shrinkwrap?!

proksch:
I found a [discussion that seems to explain how to solve the class-loading issue](https://github.com/pf4j/pf4j/issues/392). However, when I try to adopt this in our code, `FastenServer` won't find an extension anymore...

    Plugin init done: 0 KafkaPlugins, 0 DB plug-ins, 0 GraphDB plug-ins: 0 total plugins

Also upgrading to the current PF4J v3.6.0 did not help.

At this point, I don't know how to continue. My last resort would be making `PomAnalyzer` a stand-alone app, but before I do this, I will wait for @mir-am reaction to this issue, maybe he can find a way to make it work.

mir-am:
> I found a [discussion that seems to explain how to solve the class-loading issue](https://github.com/pf4j/pf4j/issues/392). However, when I try to adopt this in our code, `FastenServer` won't find an extension anymore...
> 
> ```
> Plugin init done: 0 KafkaPlugins, 0 DB plug-ins, 0 GraphDB plug-ins: 0 total plugins
> ```
> 
> Also upgrading to the current PF4J v3.6.0 did not help.
> 
> At this point, I don't know how to continue. My last resort would be making `PomAnalyzer` a stand-alone app, but before I do this, I will wait for @mir-am reaction to this issue, maybe he can find a way to make it work.

I agree that we can make POMAnalyzer a stand-alone app as a last resort. However, before doing so, I'd like to investigate the issue myself after the holidays.

MagielBruntink:
@proksch I think the `getPomUrl` method below should do it. It passes the relevant tests, at least:
```
	public String getPomUrl() {
		var localPomUri = localPomFile.toURI();
		var localM2Uri = localM2Repository.toURI();
		if (!localPomUri.getPath().startsWith(localM2Uri.getPath())) {
			var msg = "instead of local .m2 folder, file is contained in '%s'";
			throw new IllegalStateException(String.format(msg, localPomUri));
		}
		try {
			// The '/' in URIs is the correct path separator, no matter the platform.
			var repoUri = new URI(artifactRepository);
			var path =  localM2Uri.relativize(localPomUri).getPath();
			if(!repoUri.getPath().endsWith("/")) {
				path = "/" + path;
			}
			return new URL(repoUri.getScheme(), repoUri.getHost(), repoUri.getPort(), repoUri.getPath() + path).toString();
		} catch (MalformedURLException | URISyntaxException e) {
			throw new RuntimeException(e);
		}
	}
```

proksch:
On an unrelated note, I have spend some vacation-time to find a working solution for how to get the `PomAnalyzer` work and I am really happy with the proof-of-concept that I came up with. I will finalize that next week when I am back in the office and I expect that I will be able to report on it in the next dev call.

In any way, there is no need for looking into the class-loading issue of PF4J anymore...

proksch:
I assume that the remaining test error is related to the deletion/access of POMs in the user/.m2 folder. As promised, I still have it on my TODO list to look into not using the default `.m2` folder.

proksch:
After looking into the repo location for a bit, I found that it is indeed rather straight forward to set a custom `settings.xml` and move the `.m2/repository` folder:

    private static Set<ResolutionResult> resolvePom(File f) {
    	var res = new HashSet<String[]>();
    	try {
    		MavenResolvedArtifactImpl.artifactRepositories = res;
    		// this line here does the trick
    		var resolver = Maven.configureResolver().fromFile("path/to/settings.xml");
    		resolver.loadPomFromFile(f) //
    				.importCompileAndRuntimeDependencies() //
    				.resolve() //
    				.withTransitivity() //
    				.asResolvedArtifact();
    		MavenResolvedArtifactImpl.artifactRepositories = null;
    		return toResolutionResult(res);
    	} catch (IllegalArgumentException e) {
    		// no dependencies are declared, so no resolution required
    		return new HashSet<>();
    	}
    }

However, I have several reservations about this solution:

1) We can include these `settings.xml` in the resources folder, but we need two versions... one that contains a Windows path and one that has a Linux path.
2) It is not clear to me which default setting to use here? We could use relative paths (e.g., we could just use `./.m2/repository` and put it into the current working dir), but it is not really expected behavior that running a test generates new files somewhere on my disk (versus deleting a file that I will download again milliseconds later).
3) Starting Shrinkwrap with a custom `settings.xml` somehow affects Maven. At least on my machine, Eclipse shows errors on all projects afterwards, until I delete the temporary folder and refresh my workspace.

All three, but especially 3), are blockers for me, so I strongly object going that route. I have committed the code to the branch `pr372-defunct-custom-settings-for-resolver` if somebody wants to further investigate this.

I have fixed the windows build now by introducing an OS distinction. *On Windows*, resolution errors no longer cause an Exception to be thrown and instead will only be logged as an error. On other OSes, the behavior stays at it was before. As the plugin will only run in production under Linux, this will fix any inconveniences without affecting the functionality on the server.

MagielBruntink:
Hi @proksch,

I have resolved the issue with running Shrinkwrap resolution in an isolated manner from a user's local .m2 repo. This is not so hard to implemented actually, using the `maven.repo.local` System property and configuring the Shrinkwrap resolver `withClassPathResolution(false)`:
```
Maven.configureResolver() //
	.withClassPathResolution(false) //
	.loadPomFromFile(f) //
	.importCompileAndRuntimeDependencies() //
	.resolve() //
	.withTransitivity() //
	.asResolvedArtifact();
```
All that is needed further is something like:
```
tempMavenRepo = Files.createTempDirectory("pom-analyzer-resolver");
System.setProperty("maven.repo.local", tempMavenRepo.toString());

   <Production or test code>

FileUtils.deleteDirectory(tempMavenRepo.toFile());
```

I have the changes locally. How would you like to proceed?

proksch:
I'll happily check out, whether Eclipse likes that more... I used the same approach, I just set the property through settings.xml and not via a runtime environment variable.

Concern two remains unresolved though... Where do we store the files/when do we clean that folder up? A temp folder that we delete after every resolution is ok-ish, but we will have tons of unecessary downloads... Maybe wrapping the whole execution in a try/finally and deleting on program stop might be an option... What do you think?

MagielBruntink:
For tests, handle the temp directory with @BeforeAll and @AfterAll. For the actual production plugin, create the temp dir and set the system property on startup, and do a shutdown hook to remove the temp dir. The we have essentially the same level of downloads as with a default .m2 directory.

MagielBruntink:
@proksch Scratch that. Just starting the Shrinkwrap call as follows in Resolver makes the tests succeed on Windows:
```
Maven.configureResolver() //
    .withClassPathResolution(false) //
....
```
No temp directory even needed. I would suggest you just added that, and forget about the temporary dirs for now.

proksch:
I have tried the `withClassPassResolution` option and it works for me as well, I will add this to the final code.

To make the `PomAnalyzer` work, I have migrated the plugin to a new "stand-alone" setup (not quite, but more on that in our next dev call) and therefore removed it here. We can leave the PR open for now, but for the next `PomAnalyzer` release, I need to get it merged on `develop`, as the `PomAnalyzer` relies on several changes throughout the code base that were introduced in this PR along the way.

proksch:
Are there any issues left that prevent us from merging this back onto develop?

MagielBruntink:
Well, the (old) POM Analyzer will then be deleted, while the new one is not yet fully deployable (or is it)? I would really prefer to have a develop branch that we can build working containers of.

Would it be an option to keep the current POM Analyzer around as well, until it is easy to switch over?

proksch:
This is a bit of a chicken and egg problem, as the new one relies on the `core` package, which is only being released through commits to `develop`, so as long as the branch is not merged, I won't get GitHub actions (and releases) for the new one working. I see your point though... I will think about another (temporary) way to host `core` for now and will focus on polishing the last bits of the `PomAnalyzer`, we can then reconsider the merge (hopefully early) next week.

proksch:
The new loader platform is now published to https://github.com/fasten-project/data-processing 

The platform is well tested with unit-tests, has a proper release strategy for all Maven packages and the Docker container, and I have run smoke tests in the docker-compose environment to test that old and new plugins work together (I will open a PR for the changes to the `docker-compose.yml` soon).

So far, I have migrated the [pom-analyzer](https://github.com/fasten-project/data-processing/tree/main/plugins/pom-analyzer) and the [maven-crawler](https://github.com/fasten-project/data-processing/tree/main/plugins/maven-crawler) to that new infrastructure and I have paved the ground for [integration-testing](https://github.com/fasten-project/data-processing/tree/main/plugins/integration-tests).

The amount of balls that I need to keep up in the air starts to become inconveniently large, so I would appreciate, if we could finally merge this PR.

proksch:
We need this PR for the re-deployment of the pipeline. As no further comments have emerged over the past weeks, I take this as silent agreement and merge the PR onto develop.

PR #374, opened 2021-12-28
Avoid unnecessary Maven progress reports in builds

proksch:
Maven's default progress indicators make the build output hard to read. This minimal change should avoid this.

### DIFF
--- a/.github/workflows/maven_develop.yml
+++ b/.github/workflows/maven_develop.yml
@@ -24,4 +24,4 @@ jobs:
           distribution: 'temurin'
           java-version: '11'
       - name: Build with Maven
-        run: mvn clean install
+        run: mvn -B -ntp clean install
--- a/pom.xml
+++ b/pom.xml
@@ -120,6 +120,12 @@
             <version>1.7.32</version>
         </dependency>
         <dependency>
+		    <groupId>org.slf4j</groupId>
+		    <artifactId>slf4j-nop</artifactId>
+		    <version>1.7.30</version>
+		    <scope>test</scope>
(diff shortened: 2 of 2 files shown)


proksch:
To put this into perspective, this change shortens the resulting build log from 36k lines to about 10k lines. This is still way too much, but a ~75% reduction is at least a first step.

MagielBruntink:
LGTM!

proksch:
I went one step further and also included `slf4j-nop` as the default backend for SLF4J in the test scope. This cuts the build log further down by 1/3 to about 3.5k lines and makes it actually possible to read it.

Unfortunately, we will have to live with a warning in the build that multiple SLF4J bindings can be found in the class, but I cannot change that as many of our projects and also some of our direct dependencies directly depend on `logback-classic`, which includes a binding out-of-the box, so it cannot be excluded. :/ To make things worse, some projects even directly work with logback classes, which prevent any kind of high-level management, because it solidifies the dependency to logback. I am not going to talk about all the sysouts that we have in the code everywhere.

I think at some point we should unify our logging practices in the project... as far as I understood, the idea is to only code against the `slf4j-api` dependency and have an `slf4j-nop` binding for the testing scope to keep the terminal output clean (= non-existing). This dependency can even be excluded by dependents if necessary. I do not remember where I found that rule in the SLF4J docs, but one should include an actual logging backend like logback or any other SLF4J implementation *only in projects that are being executed*, otherwise, personal logging preferences will be forced up all dependents.

mir-am (approved):
LGTM as well.

PR #376, opened 2021-12-28
Reduce the logs generated by merger and cg generator

ashkboos:
## Description
Remove some of the less necessary yet frequent log statements from the merger and CG generator not to pollute our builds.


proksch:
I am not objecting if you remove unnecessary statements, but the cases that you have addressed here should already be solved in the recent PR #374 , which introduces `slf4j-nop` as the logging backend during the build, which will disable all logging that goes through SLF4J.

The more important change would be to identify locations in the code that use `sysout` to print something during the build, because this cannot be suppressed by simply switching the logging backend.

I would suggest to execute the build on develop with the recent changes. All logging statements that still pop-up are not done via SLF4J and these are the ones that should be addressed. They do not even have to be removed, just changing them to SLF4J logger calls would be enough.

proksch:
@ashkboos This PR is still open, I am just wondering, what is the reason... Do you want to perform further changes or did it just slip through? Please consider my last comment regarding SLF4J logging, but regardless whether you decide to change the logging statements to SLF4J or remove the log statements (I do not have a strong opinion there, just providing explanation), please rebase once more and resolve the current conflicts, then you can just merge it into develop.

PR #377, opened 2022-01-02
Quickfix: package search query for Rest API

romatallinn:
## Description
The PR is a quick fix for the implementation of the search query that is used in Rest API. 

## Motivation
The search API endpoint is used by [fasten-project/fasten-web](http://github.com/fasten-project/fasten-web), so the users can search the packages conveniently on the web interface. Due to the problem in the query, this functionality is currently unavailable.


mir-am (approved):
LGTM!

PR #362, opened 2021-12-07
Integrate Vulnerable Call Chains Repository interface in Rest API

romatallinn:
## Description
The PR adds API endpoints that implement Vulnerable Call Chains Repository.

## Motivation and context
[fasten-project/fasten-web](https://github.com/fasten-project/fasten-web/) needs to retrieve Vulnerable Call Chains from the server to nicely show the results to the user in the web application.

PR #378, opened 2022-01-02
Integrate Vulnerable Call Chains Repository interface in Rest API

romatallinn:
## Description
The PR adds API endpoints that implement Vulnerable Call Chains Repository.

## Motivation and context
[fasten-project/fasten-web](https://github.com/fasten-project/fasten-web/) needs to retrieve Vulnerable Call Chains from the server to nicely show the results to the user in the web application.

## Additional context
Cherry-picked clean version of #362.

### DIFF
--- a/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/VulnerableCallChainsApiServiceImpl.java
+++ b/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/VulnerableCallChainsApiServiceImpl.java
@@ -21,11 +21,10 @@ package eu.fasten.analyzer.restapiplugin.mvn.api.impl;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import eu.fasten.analyzer.restapiplugin.mvn.KnowledgeBaseConnector;
-import eu.fasten.analyzer.restapiplugin.mvn.api.VulnerabilityCallChainApiService;
+import eu.fasten.analyzer.restapiplugin.mvn.api.VulnerableCallChainsApiService;
 import eu.fasten.core.data.FastenURI;
 import eu.fasten.core.vulchains.VulnerableCallChain;
 import eu.fasten.core.vulchains.VulnerableCallChainRepository;
-import org.json.JSONArray;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.annotation.Lazy;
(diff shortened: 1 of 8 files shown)


romatallinn:
@ashkboos @proksch the PR is ready to be merged. Would appreciate a review or a sign that I can merge it.

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/VulnerableCallChainsApiServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    /**
+     * Helper method to convert set of vulnerable call chains to JSON formatted string response.
+     * @param chains - a set of {@link VulnerableCallChain} objects to be serialized.
+     * @return {@link String} formatted as JSON response.
+     */
+    private String VulnerableCallChainsToJSON(Set<VulnerableCallChain> chains) {
+        GsonBuilder builder = new GsonBuilder();
```
ashkboos, line 62: We have `eu.fasten.core.vulchains.VulnerableCallChainJsonUtils` that you can use for the serialization and deserialization. I would suggest using that for the sake of consistency.

ashkboos (commented):
It looks good. I have one suggestion that I would appreciate if you do before merging the PR. I have collected a good number of real-world examples in `mnt/fasten/vul-path`. Could you point it to this location so we can see some examples on the website for real?

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/impl/VulnerableCallChainsApiServiceImpl.java
romatallinn: > mnt/fasten/vul-path

Sure. It's more of a change for k8s-deployments repo. I changed it. You may also want to have a look at PR there.

> eu.fasten.core.vulchains.VulnerableCallChainJsonUtils

Well, the format you store as json in your files is different what we want to return to client. Do you remember the thing about different types of fasten uri? So, it still doesn't work as expected. It still returns just a raw uri in json, and not the object with parsed properties.

Although I don't think you need it such in your implementations. So maybe we can just to create a different json util for my case, rather than adjusting yours?

romatallinn: So, to remind you the difference,

Screenshot 2022-01-17 at 12 59 03
Screenshot 2022-01-17 at 13 00 16

The first is what I need on the client, the second is what returned by your json util.

ashkboos (approved):

Issue #313, opened 2021-10-12
Bug in the URI of Python Extended Revision Call Graphs


gdrosos:
## Describe the bug
When running the Metadata Plugin for the Python Database, for each Python product, an Extended Revision Python Call Graph is generated. This Graph contains a specific attribute, namely Uri, which includes the forge, product and version of the entity, separated by an exclamation mark and a dollar sign:
https://github.com/fasten-project/fasten/blob/3982b5656543d4c1ae9c93ad6fa87256cca63cbb/core/src/main/java/eu/fasten/core/data/ExtendedRevisionCallGraph.java#L170 Some Python packages, like pulp-smash, include the  exclamation mark in their version name, for example version: 1!0.0.7. 
When the FastenURI is constructed for versions like the aforementioned, the code falsely assumes that the URI includes multiple forges, because the product version includes exclamation marks. 
https://github.com/fasten-project/fasten/blob/3982b5656543d4c1ae9c93ad6fa87256cca63cbb/core/src/main/java/eu/fasten/core/data/FastenURI.java#L54-L63 As a result, an IllegalArgumentException erroneously occurs at Runtime.
## To Reproduce
1. Run the MetadataDBPythonExtension Plugin in the server:
 ```FASTEN_DBPASS=pass java -cp docker/server/server-0.0.1-SNAPSHOT-with-dependencies.jar eu.fasten.server.FastenServer  -p "docker/plugins" -pl "MetadataDBPythonExtension" -d  "PyPI=jdbc:postgresql://fasten@address/fasten_python" -k localhost:9092  -kt "MetadataDBPythonExtension=fasten.pycg.cvt.out"``` 

## Expected behavior
The Python Metadata Plugin is expected to run properly, without any errors or exceptions

## Logs
If applicable, add logs to help explain your problem (e.g. the error message and/or exception traceback).
```
[2021-10-12 13:18:37,875] [INFO ] [main] [e.f.s.p.k.FastenKafkaPlugin] - Read message offset 559 from partition 62.
[2021-10-12 13:18:37,879] [ERROR] [main] [e.f.s.p.k.FastenKafkaPlugin] - Error occurred while processing call graphs
java.lang.IllegalArgumentException: More than one forge
        at eu.fasten.core.data.FastenURI.<init>(FastenURI.java:63)
        at eu.fasten.core.data.FastenURI.create(FastenURI.java:118)
[...]

PR #381, opened 2022-01-05
Fix bug in the Fasten URI 

gdrosos:
## Description
Currently the Fasten Uri (forge!product$version) does not support package versions containing the exclamation mark (like  [palpsmash:1!0.14.0](https://pypi.org/project/pulp-smash/)).  
This happens because  the code falsely assumes that the URI contains multiple forges. 
With this PR, we perform all the corresponding forge checks only to the URI substring of forge and package, so that the check will not be affected by an exclamation mark in the version.

## Motivation and context
Why is this change required? What problem does it solve?
If it fixes an open issue, please link to the issue here.

This Pull Request aims to fix the issue described on #313

## Testing
Please describe in detail how you tested your changes.
* Adapted the FastenURITests to support the new valid FastenURI format.
* Now the Metadata Plugin can successfully store those python revisions to the metadata database, where previously it throwed an IllegalArgumentException, as described on #313


romatallinn:
LGTM.

ashkboos:
@vigna what do you think about the changes in this PR? Can we allow `!` in the version? Do you know any better option? I think encoding the version could be an option.

vigna (approved):
This seems to be fine. There is no ambiguity in having `!` in the version as it has to follow a product, so there's a `$` before.

PR #382, opened 2022-01-07
Fix ExtendendGID Graphs produced by C & Python Metadata Plugin

gdrosos:
## Description
This Pull Request aims to fix the ExtendendGID Graphs produced by C & Python Metadata Plugin on two ways:

### Feed the C output topic with ExtendedGID graphs instead of GID
Firstly, regarding the C Metadata Plugin, it produced graphs following the GID format  for the output kafka topic . 
In order to be properly employed by the Callable Indexer and thus stored in the RocksDao db,
 the produced C Graphs were modified to follow the ExtendedGIDFormat, just like Python and Java.

### Avoid NullPointerException when producing ExtendedGID Graphs.
Secondly, as pointed out by @MagielBruntink, despite the fact that the Python Metadata Plugin succesfully stored metadata in the db, it failed to produce the ExtendedGID Graphs to the output topic. This happened because as can be seen bellow:

https://github.com/fasten-project/fasten/blob/600985190e07aa479a6f25bd2e13155dc084c941/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabasePythonPlugin.java#L167

https://github.com/fasten-project/fasten/blob/600985190e07aa479a6f25bd2e13155dc084c941/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDatabaseCPlugin.java#L247

for both C & Python only the Graph edges were stored in the CallSites table. This resulted on an NullPointerException later, as in order to get a json representation of the ExtendedGID Graph, the attributes of the Call Sites table where retrieved from the db on the following way:

https://github.com/fasten-project/fasten/blob/600985190e07aa479a6f25bd2e13155dc084c941/core/src/main/java/eu/fasten/core/data/callableindex/ExtendedGidGraph.java#L80-L86

This resulted on an NullPointerException, preventing the production to the output topic. 
This exception was not triggered for java revisions, meaning that in the java db, all the call sites table attributes  were always popullated, avoiding every time the exception.

Thus, in order to tackle this issue, since Java always has populated attributes whereas C & Python never, we created a new toJson() method which is the exact same with the initial one, with the main difference that  we populate all the "callsites_information" attributes of the ExtendedGID Graph with null values, in order to follow the exact same format with Java. This method is employed when processing C & Python Revisions on the Metadata Plugin.

## Motivation and context

With this fix, the Metadata C & Python Plugin will have the excepcted behaviour. instead of triggering an NullPointerException.

## Testing

I tested both C & Python Metadata Plugins by feeding on them Dummy Kafka Topics, and observing the results on the produced topics( fasten.MetadaDBC/PythonExtension.err & fasten.MetadaDBC/PythonExtension.out). The Graphs had the expected format without  producing any record in the Error Kafka topics.

### DIFF
--- a/core/src/main/java/eu/fasten/core/data/callableindex/ExtendedGidGraph.java
+++ b/core/src/main/java/eu/fasten/core/data/callableindex/ExtendedGidGraph.java
@@ -75,14 +75,25 @@ public class ExtendedGidGraph extends GidGraph {
     public JSONObject toJSON() {
         var json = super.toJSON();
         var callSitesInfo = new JSONObject();
-        getCallsInfo().forEach((edge, info) -> {
-            var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
-            var infoJson = new JSONObject();
-            infoJson.put("line", info.getLine());
-            infoJson.put("receiver_type_ids", new JSONArray(Arrays.asList(info.getReceiverTypeIds())));
-            infoJson.put("call_type", info.getCallType().getLiteral());
-            callSitesInfo.put(edgeStr, infoJson);
-        });
+        try {
+            getCallsInfo().forEach((edge, info) -> {
+                var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
+                var infoJson = new JSONObject();
+                infoJson.put("line", info.getLine());
+                infoJson.put("receiver_type_ids", new JSONArray(Arrays.asList(info.getReceiverTypeIds())));
+                infoJson.put("call_type", info.getCallType().getLiteral());
+                callSitesInfo.put(edgeStr, infoJson);
+            });
+        } catch(NullPointerException e) {
+            getCallsInfo().forEach((edge, info) -> {
+                var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
+                var infoJson = new JSONObject();
+                infoJson.put("line", JSONObject.NULL);
+                infoJson.put("receiver_type_ids", JSONObject.NULL);
+                infoJson.put("call_type", JSONObject.NULL);
+                callSitesInfo.put(edgeStr, infoJson);
+            });
+        }
         json.put("callsites_info", callSitesInfo);
         var gidToUriJson = new JSONObject();
         this.gidToUriMap.forEach((k, v) -> gidToUriJson.put(String.valueOf(k), v));
--- a/core/src/main/java/eu/fasten/core/data/callableindex/ExtendedGidGraph.java
+++ b/core/src/main/java/eu/fasten/core/data/callableindex/ExtendedGidGraph.java
@@ -71,29 +71,39 @@ public class ExtendedGidGraph extends GidGraph {
         return typeMap;
     }
 
+    public JSONObject toCPythonJSON() {
+        var json = super.toJSON();
+        var callSitesInfo = new JSONObject();
+        getCallsInfo().forEach((edge, info) -> {
+            var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
+            var infoJson = new JSONObject();
+            infoJson.put("line", JSONObject.NULL);
+            infoJson.put("receiver_type_ids", JSONObject.NULL);
+            infoJson.put("call_type", JSONObject.NULL);
+            callSitesInfo.put(edgeStr, infoJson);
+        });
+        json.put("callsites_info", callSitesInfo);
+        var gidToUriJson = new JSONObject();
+        this.gidToUriMap.forEach((k, v) -> gidToUriJson.put(String.valueOf(k), v));
+        json.put("gid_to_uri", gidToUriJson);
+        var typesJson = new JSONObject();
+        this.typeMap.forEach((k, v) -> typesJson.put(String.valueOf(k), v));
+        json.put("types_map", typesJson);
+        return json;
+    }
+
     @Override
     public JSONObject toJSON() {
         var json = super.toJSON();
         var callSitesInfo = new JSONObject();
-        try {
-            getCallsInfo().forEach((edge, info) -> {
-                var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
-                var infoJson = new JSONObject();
-                infoJson.put("line", info.getLine());
-                infoJson.put("receiver_type_ids", new JSONArray(Arrays.asList(info.getReceiverTypeIds())));
-                infoJson.put("call_type", info.getCallType().getLiteral());
-                callSitesInfo.put(edgeStr, infoJson);
-            });
-        } catch(NullPointerException e) {
-            getCallsInfo().forEach((edge, info) -> {
-                var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
-                var infoJson = new JSONObject();
-                infoJson.put("line", JSONObject.NULL);
-                infoJson.put("receiver_type_ids", JSONObject.NULL);
-                infoJson.put("call_type", JSONObject.NULL);
-                callSitesInfo.put(edgeStr, infoJson);
-            });
-        }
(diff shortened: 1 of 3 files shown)


core/src/main/java/eu/fasten/core/data/callableindex/ExtendedGidGraph.java
```diff
@@ -75,14 +75,25 @@ public ExtendedGidGraph(long index, String product, String version, List<Long> n
...
         var callSitesInfo = new JSONObject();
-        getCallsInfo().forEach((edge, info) -> {
-            var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
-            var infoJson = new JSONObject();
-            infoJson.put("line", info.getLine());
-            infoJson.put("receiver_type_ids", new JSONArray(Arrays.asList(info.getReceiverTypeIds())));
-            infoJson.put("call_type", info.getCallType().getLiteral());
-            callSitesInfo.put(edgeStr, infoJson);
-        });
+        try {
```
MagielBruntink, line 78: Thanks for looking into a fix. I'm not too happy with the chosen way to fix it, however. These null pointer exceptions will happen for each C and Python graph. By catching and then fixing the issue, I feel we are dealing with this too late. Exceptions should really be reserved for unexpected situations. Can we fix this earlier in the code?

gdrosos: Thanks @MagielBruntink  for the immediate review, I changed the approach. More specifically I created a separated toJson() method for C & Python packages, which is the exact same with the toJson() method but instead of acessing null values in Call sites records, it just inserts null values to the callsites_info Json attributes. This method is employed when thei C & Python metadata Plugins are employed. Now the Exception will not be triggered

MagielBruntink (changes requested):
Hi @gdrosos,

I would prefer solution like this, so that we don't make the generic MetadataPlugin and ExtendedGIDGraph classes more specific to C, Java, and Python:

In ExtendedGIDGraph the `toJSON` method should just check if the edge info object is null:
```
    @Override
    public JSONObject toJSON() {
        var json = super.toJSON();
        var callSitesInfo = new JSONObject();
        getCallsInfo().forEach((edge, info) -> {
            var edgeStr = String.format("[%d, %d]", edge.getFirst(), edge.getSecond());
            var infoJson = new JSONObject();
            infoJson.put("line", info != null ? info.getLine() : JSONObject.NULL);
            infoJson.put("receiver_type_ids", info != null ? new JSONArray(Arrays.asList(info.getReceiverTypeIds())) : JSONObject.NULL);
            infoJson.put("call_type", info != null ? info.getCallType().getLiteral() : JSONObject.NULL);
            callSitesInfo.put(edgeStr, infoJson);
        });
        json.put("callsites_info", callSitesInfo);
        var gidToUriJson = new JSONObject();
        this.gidToUriMap.forEach((k, v) -> gidToUriJson.put(String.valueOf(k), v));
        json.put("gid_to_uri", gidToUriJson);
        var typesJson = new JSONObject();
        this.typeMap.forEach((k, v) -> typesJson.put(String.valueOf(k), v));
        json.put("types_map", typesJson);
        return json;
    }
```

So then the specific toCPythonJSON method is not needed, and you don't need to have this switch in MetaDataPlugin either:
```
    @Override
    public Optional<String> produce() {
        if (gidGraph == null) {
            return Optional.empty();
        } else {
            if (this.forge.equals(Constants.mvnForge)) {
                return Optional.of(gidGraph.toJSON().toString());
            } else {
                return Optional.of(gidGraph.toCPythonJSON().toString());
            }
        }
    }
```

gdrosos:
Hello @MagielBruntink , thanks for the proposed sollution. Actually this proposal was my initial attempt to resolve this issue, but the edge info object (The CallSitesRecord) is not Null because the CallSitesRecord attributes SourceId and TargetId contain a value for both C & Python. If we try this approach, we still get a NullPointerException. In order to implement a generic sollution for all languages, we need to be able to check if the CallSitesRecord attributes are null without calling the  CallSitesRecord getters(as they trigger the Exception), but I do not know how we can achieve this given the current CallSitesRecord methods generated by JOOQ. That is why I thought the best way to avoid the Exception is by implementing a new method for C & Python.

BTW the callsites_info attribute is not used in the CallableIndexer Plugin (which is the one consuming the MetadataDB out topic), so we can just delete this attribute since it is not employed in any part of the pipeline.

MagielBruntink:
Hi George, I see. I would like to have @mir-am and @ashkboos take a look here - I don't know why the getters throw NPEs, I think it points to an underlying issue.

mir-am:
> 



> Hi George, I see. I would like to have @mir-am and @ashkboos take a look here - I don't know why the getters throw NPEs, I think it points to an underlying issue.

I'm currently on holiday till the end of the week. I'll take a look at the issues from next Monday.

ashkboos:
> Hi George, I see. I would like to have @mir-am and @ashkboos take a look here - I don't know why the getters throw NPEs, I think it points to an underlying issue.

Hi Georgios and Magiel,
I looked into it. I am not closely familiar with this part of code but my understanding is that NPE could be due to the following constructor call. Because by calling `info.getCallType().getLiteral()` getCallType returns null and getLiteral result in npe. 
image
I think a proper solution is to prevent having the same object of callSiteInfo for all three langs. But even if we don't want to implement separate classes I think in the constructor call we should at least provide empty objects of needed type instead of null.

gdrosos:
> > Hi George, I see. I would like to have @mir-am and @ashkboos take a look here - I don't know why the getters throw NPEs, I think it points to an underlying issue.
> 
> Hi Georgios and Magiel, I looked into it. I am not closely familiar with this part of code but my understanding is that NPE could be due to the following constructor call. Because by calling `info.getCallType().getLiteral()` getCallType returns null and getLiteral result in npe. image I think a proper solution is to prevent having the same object of callSiteInfo for all three langs. But even if we don't want to implement separate classes I think in the constructor call we should at least provide empty objects of needed type instead of null.

Hello, indeed this is the root cause of the issue, because I tested calling the getSourceId and getTargetId methods and they do not trigger any NPEs for either C or Python, so the getters of the null attributes cause the NPE.

gdrosos:
Hello, I tried to provide empty objects of needed type of each null attribute ( succesfully 0 for line and an empty Long array for RecieverTypeIds) but I cannot create an empty Value of a CallType because it is an enum type auto generated by JOOQ:
https://github.com/fasten-project/fasten/blob/4ae9a5743c2262dad4f1eb95e961835f0a6027c8/core/src/main/java/eu/fasten/core/data/metadatadb/codegen/enums/CallType.java#L27-L42
from this SQL command:
``` CREATE TYPE CALL_TYPE AS ENUM ('static', 'dynamic', 'virtual', 'interface', 'special');```

If we want to add an empty value we should add an extra "empty" type to this enum. Otherwise, since for C and Python we do not have a Calltype, I believe it is better to have a separate toJson method for the the production of the ExtendedGidGraphs. What do you think? @MagielBruntink  @ashkboos

MagielBruntink:
Hi @gdrosos it seems we have two options for now:
 1. the solution you proposed before by adding a specific toJson method and a switch somewhere else. It will increase the complexity of the existing classes, but it relatively easy to do now.
 2. slightly adapt the SQL & JOOQ code to extend the enum as you described, and then re-writing parts of the code to use that.

I think we should do with 1. for now and at the same time create an issue that requests the revision of these classes - it seems to me there are more suboptimal choices in them. The most important consideration for me is that we need to get the C and Python pipelines up and running now.

MagielBruntink (approved):

MagielBruntink:
I'm fine. @ashkboos ?

ashkboos (approved):
Let's merge this, just to make things work for now. I think we should do the similar refactoring that we did in PR #369, this time for MetadataDBExtension and ExtendedGID classes and all other classes that are related to them. But that should be in another PR.

PR #388, opened 2022-01-28
Fix vulnerability statements processor db cleanup

MagielBruntink:
## Description
Improved the vulnerability-statements-processor by cleaning up DBs of all ecosystems. This is useful in case old data is remaining after bugs or data issues.


proksch (approved):
LGTM

PR #395, opened 2022-01-31
Remove cache plugin

romatallinn:
## Description
Removes depricated notices of Vulnerability Cache processor.

## Motivation and context
With the implementation of the new Vulnerable Call Chains Repository and its implementation in Rest API, we still have similiar logics for Vulnerability Cache processor and vulnerabilities endpoint. In order not to have confusing tools, this PR removes depricated ones from the production.


MagielBruntink:
@romatallinn For the cache invalidation, there are two plugins, a "normal" one and a parallel one. Are both necessary?

romatallinn:
@MagielBruntink the "normal" one consumes from callable index extension, the "parallel" one consumes from vulnerability consumer extension. This ensures that the cache is updated when (I) a new artifact has been process by the pipeline, (II) the vulnerability information is updated for the artifact. Otherwise, we can miss some important events that affect consistency of data.

And due to Kafka limitation, one plugin cannot consume from two Kafka topics simultaneously. We did look into alternative ways and workarounds. But as I said, we are still discussing how better to integrate these plugins into new Vulnerable Call Chains system.

mir-am (approved):
LGTM!

PR #396, opened 2022-01-31
Update Metadata DB schema for ingested artifacts

proksch:
This PR proposes to change the ingestion table to only contain a generic `key` that can be checked for existence. This allows for a more generic handling of the duplicate detection and can be used as the respective plugin sees fit. For example, Java's PomAnalyzer will use keys like `gid:aid:packaging:version-lane`, but C/Python might want to use different keys, e.g., if they do not use the lane concept. We could also think about using this key for additional duplication checks later in the pipeline, e.g., by making the pipeline stage part of it.

The adopted schema file can be found in the [docker-compose repository](https://github.com/fasten-project/fasten-docker-deployment/pull/23).


analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/LazyIngestionProvider.java
```diff
@@ -60,7 +67,7 @@ public static void ingestArtifactIfNecessary(String packageName, String version,
                 jsonRecord.put("date", date);
             }
             Timestamp curTime = new Timestamp(System.currentTimeMillis());
-			KnowledgeBaseConnector.kbDao.insertIngestedArtifact(packageName, version, curTime);
+            KnowledgeBaseConnector.kbDao.insertIngestedArtifact(toKey(packageName, version), "0.0.1-SNAPSHOT");
```
mir-am, line 70: Why do we have `0.0.1-SNAPSHOT`?

mir-am (approved):
Overall, LGTM.

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/LazyIngestionProvider.java
proksch: We do not have a proper versioning strategy and `0.0.1-SNAPSHOT` is used as the version in all poms of the old server, so I felt it is appropriate to use it here :) The main user of said methods is the new `PomAnalyzer`, which will provide proper versions.

The concrete value is also not that important and might only become interesting in the future. The intention was to give us the means to void ingestions that have been done with a particular version, should we detect a bug in said version. What we will be probably be doing is to just wipe the table every now and then, so the version might be irrelevant.

PR #397, opened 2022-01-31
Fix processing slowdown when messages only arrive in normal lane

proksch:
Right now, the server will slow down for messages on normal, as it will always wait on the PRIO lane. This PR changes this: we ONLY wait on PRIO and ONLY when there have been no messages in any lane on the previous poll cycle. The only corner case is when are no subscription for PRIO, then we wait in NORMAL instead.

Waiting is necessary as Duration.ZERO does not seem to trigger any re-assign/rebalancing of the topics in case consumers change and also to avoid contant polling in times when no new messages arrive.

(see https://github.com/fasten-project/fasten-docker-deployment/pull/22 for context)

### DIFF
--- a/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
+++ b/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
@@ -59,7 +59,7 @@ import eu.fasten.server.plugins.FastenServerPlugin;
 
 public class FastenKafkaPlugin implements FastenServerPlugin {
 
-    private static final Duration POLL_TIMEOUT = Duration.ofMillis(250);
+    private static final Duration POLL_TIMEOUT = Duration.ofSeconds(2);
 
     private final Logger logger = LoggerFactory.getLogger(FastenKafkaPlugin.class);
     private final KafkaPlugin plugin;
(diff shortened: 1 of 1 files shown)


proksch:
The weird behavior that [has been reported](https://github.com/fasten-project/fasten-docker-deployment/pull/22) (normal messages are not consumed after PRIO finished) could be attributed to a simple logic mistake in when to skip normal processing. The fix has been a one liner.

ashkboos (approved):

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -220,28 +222,40 @@ public void handleConsuming() {
         // processed
         sendHeartBeat(connNorm);
 
+        // normally, we ONLY wait on PRIO and ONLY when there have been no messages in any lane...
+        var prioTimeout = hadMessagesOnLastPollCycle ? Duration.ZERO : POLL_TIMEOUT;
```
mir-am, line 226: @proksch 
Why poll timeouts are different? Could you please elaborate a bit?

proksch: The idea is that -in general- we do not wait as long as there are messages to process (Duration.ZERO), but when there have been no messages whatsoever in the last cycle, neither on NORMAL nor on PRIO, we wait a bit... otherwise, we would end up sending constant requests to Kafka.

proksch: If you would always wait for a fixed wait time (as it was before), then Lane A will slow down Lane B when there are no messages on A, but messages on B (You can use any combination of NORMAL/PRIO for this example).

Issue #138, opened 2020-11-25
The id generated for a Java class without any package contains "null"


tmortagne:
## To Reproduce
Analyze a java class `MyClass` with no package: you end up with a JSON containing:

```
"internalTypes": {"/null/MyClass": {
```

in the produced JSON.

## Expected behavior

I feel that en empty string (so "//MyClass") would make more sense than "null".

proksch:
I would argue that the correct type for classes in the default package should be `/MyClass`.

@ashkboos The issue report is quite old by now... how has the handling been changed in the meantime?

proksch:
Please make sure that this is not forgotten... please also consider #248 and #260 as all of them are related to `FastenURI`s.

ashkboos:
@proksch I thought about these issues a lot. All of them are somehow related to the FastenUris design. We can perhaps spend some time and do some workarounds to hide them but I'm not sure it is worth the time. I think a proper solution is as you mentioned an ANTLR-based solution or sth similar.

Issue #260, opened 2021-03-26
FastenUriUtils parsing incorrectly the "class", "method" for some cases with lambda function.


cg122:
## To Reproduce

Parsing the partial uri in the following FASTEN uri

fasten://mvn!nl.tudelft.jpacman:jpacman-framework$8.1.1/nl.tudelft.jpacman.ui/PacManUiBuilder$addStopButton%28Lnl$tudelft$jpacman$game$Game%3A%29V%3A30$Lambda.$newInstance(%2Fnl.tudelft.jpacman.game%2FGame)PacManUiBuilder$addStopButton%28Lnl$tudelft$jpacman$game$Game%3A%29V%3A30$Lambda

output:

[nl.tudelft.jpacman.ui, PacManUiBuilder$addStopButton(Lnl$tudelft$jpacman$game$Game:)V:30$Lambda, ui/PacManUiBuilder$addStopButton, Lnl$tudelft$jpacman$game$Game:)V:30$Lambda.$newInstance(/nl.tudelft.jpacman.game/Game)PacManUiBuilder$addStopButton(Lnl$tudelft$jpacman$game$Game:, V:30$Lambda.$newInstance(/nl.tudelft.jpacman.game/Game)PacManUiBuilder$addStopButton(Lnl$tudelft$jpacman$game$Game:)V:30$Lambda]

## Expected behavior

class name should be "PacManUiBuilder", method should be "addStopButton"

## Additional context
Result got on develop branch (1001e75ac0a45d2126b7f33b767d9ecee9ef3364)

romatallinn:
@cg122 I adjusted implementation towards the better way. It returns `PacManUiBuilder` as the class, but `newInstance` as the method. However, it is more of an "ideological problem" which of two to return (`newInstance` or `addStopButton `).

@ashkboos will contact you to know some context behind it. Do you think there might be different scenarios when both of them are needed? In this case, we need to solve this issue somehow? Maybe an array of method names in the order of inclusion?

@cg122 also, do you maybe have artifacts similar to this one, with inner classes? Just to test the implementation.

cg122:
Thanks @romatallinn. I will look at it and see how it works for other example.

cg122:
I checked the source code corresponding to the FASTEN uri above:
```
    private void addStopButton(final Game game) {
        assert game != null;

        buttons.put(STOP_CAPTION, game::stop);
    }
```
[...]

Issue #248, opened 2021-02-25
The format of the signature in the eu.fasten.core.data.JavaNode is very strange


tmortagne:
It looks like `method()/some.package/ReturnType` while I would have expected something closer to what we usually see in Java like "some.package.method()" or "some.package#method()".

That being said it feels wrong to store this full String instead of separated elements which are only merged when you call `getSignature()` for example. That way it's a lot easier to create whatever format you want and more importantly it reduces drastically the current huge duplication this storage cause in memory because then you can properly reuse the same String instance for the package (and even imagine a list of strings instead of a single String for the package to reduce even more the duplication), the class, the return types, etc.

Regarding the return type, while it's definitely useful information for some use cases it's not part of the method signature (it's part of the declaration along with other erased stuff like the method parameters annotations and generics) and should probably be moved to the metadata to avoid polluting the signature and, worst, probably cause problems during stitching when you try to match an external callable from one package to an internal callable from another during stitching for example when the return type slightly changes from one version to another of some dependency.

proksch:
Not sure if this issue still applies, but I think the `FastenURI` syntax is baked deeply into the project and hard to remove. The silver lining is that it should not be necessary to work with these String. The FastenURI class should take care of the parsign and provide access to all contained information.

ashkboos:
There are two reasons for this 1) Unfortunately in some cases, JVM and Java code does not produce 1:1 methods. IIRC it was for generic types that JVM would generate synthetic methods. The only way that we could capture everything from bytecode was to include return types in the signature. 2) As Sebastian mentioned `FastenUri`. 
However, I agree that our representation is a bit inefficient. Some of it is caused by the fact that we initially wanted a generic solution for all 3 languages (e.g. FastenUri). Now that we know our pipeline is quite independent for each language, we are revising our steps accordingly. I think in one of our next refactorings where we refactor CG generation and merging, we need to revise our representation drastically, we might even be able to use the serializer/deserializer provided by OPAL itself if we do not need a generic solution. Then such problems will be solved automatically.

proksch:
@ashkboos Could you please follow up on this ticket or close it once it is resolved?

Anther related ticket is #260 which also points out an issue with the `FastenUri` creation.

PR #398, opened 2022-01-31
using empty namespace for types with no package

ashkboos:
As suggested in issue #138 we now use empty namespace in the FastenUris of types without package instead of using null


analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/CGandStitchingTest.java
```diff
@@ -102,6 +103,12 @@ public void staticInitializer() throws OPALException, URISyntaxException {
             "\n", toString(mergedRcg));
     }
 
+    @Test
+    public void noPackage() throws URISyntaxException {
+        final var np = getRCG("merge/staticInitializer/NoPackage.class", "noPackage", "0.0.0");
```
proksch, line 108: Shouldn't this be "NoPackage"? (capital case N)

analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/analysis/OPALMethod.java
```diff
@@ -104,7 +107,7 @@ public static String getPackageName(final Type parameter) {
                 return slashToDot(Objects
                         .requireNonNull(getPackageName(parameter.asArrayType().componentType())));
             } else if (parameter.asObjectType().packageName().equals("")) {
-                return null;
+                return "";
```
proksch, line 110: Do we then end up with a double slash in the URLs? So before it was "/null/ClassName", now it is "//ClassName"... or is this being handled through canonicalize?

proksch (commented):
I have two minor clarification questions...

analyzer/javacg-opal/src/test/java/eu/fasten/analyzer/javacgopal/CGandStitchingTest.java
ashkboos: fixed

analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/analysis/OPALMethod.java
ashkboos: Yes, now it is "//ClassName". Canonicalize does not handle it. In fact, if there is no namespace canonicalize complains about it. That is why we now check before canonicalizing to see if there is not any namespace please don't try.

proksch: Hmm, I do not like the `//` construct... I would argue that the correct version would be `/ClassName`. Is there an easy way to achieve this?

ashkboos: It is not allowd by the FastenUri class if you write this: 
```
var v = FastenURI.create("/ClassName");
```
you would get `java.lang.IllegalArgumentException: Missing entity`

ashkboos: I also just realized that even "//ClassName" cannot be correctly parsed by FastenUri. It does not cause any exception but It assumes that `ClassName` is the product name while it is the name of the class. What should we do about this? should we use some meaningful placeholder for the package name such as `/no.package/ClassName`? or should we create another URI specific for the classes?

proksch: How about using `/ClassName` and fixing the code that extracts the package/product/class information?

proksch: In general, I think we should revise the `FastenURI` at some point. We should improve our validation and I also think that it does not make sense to parse the whole thing on initialization. I would prefer storing the string representation and only parse part of it on demand, we can still cache information, once parsed. ... but this is a discussion for another issue.

proksch (changes requested):
It seems that we should introduce proper handling for the problem at hand as an ad-hoc solution does not seem to solve the problem. I propose that we move this to the next release.

mir-am:
With Mehdi, we decided to close this PR. Because we should do a major refactoring on the FASTEN URIs or use another technique.

amottier:
@mir-am @ashkboos did you open an issue to track discussion on the "major refactoring on the FASTEN URIs" mentioned in the @mir-am comment? I guess that a refactoring would address issue #138, #248 and #260. Also can you remind me in which deliverable current implementation of URI is specified? Thanks.

PR #404, opened 2022-02-08
Simplify Rest-API

gdrosos:
## Description
Simplified the REST-API Plugin structure by removing all the API Service and API Service Implementation classes, and moving the implementation directly to the initial API corresponding class. 

## Motivation and context
Why is this change required? What problem does it solve?
If it fixes an open issue, please link to the issue here.

Currently, the REST-API plugin was unnecessarily complex. 
To be more precise, the current implementation included for each endpoint one API class where the mapping of the methods was specified, one API Service class were an interface with its methods was definied and one corresponding API Service Implementation class were the implementation of each method was taking place. 
This structure was unnecessarily complex, and after a discussion with @proksch, we decided to reduce the complexity of those classes by merging them into one.
To adduce an example, the PackageApi, PackageApiService and PackageApiServiceIml classes have been merged into a single PackageApi class.

## Testing
Please describe in detail how you tested your changes.
* Removed all the serivce tests and kept only the previous implementation tests, which now test directly their corresponding API method.
* Tested all the endpoints locally and they have the expected behaviour


mir-am (approved):
LGTM!

PR #405, opened 2022-02-09
Quickfix: Add CLI args to the DependencyGraphBuilder main method

### LINKED ISSUE #304: Remove the hard-coded DB credentials from the Maven Dependency Graph Code
We should not reveal our DB URL and username in the code!!!

https://github.com/fasten-project/fasten/blob/941d8fc194476d5dcb719a5c5b4c69125e3b77c6/core/src/main/java/eu/fasten/core/maven/DependencyGraphBuilder.java#L52

gdrosos:
## Description
Added  command line arguments when triggering DependencyGraphBuilder main method, in order to initialize the class in a secure way.

## Motivation and context

The Dependency Graph Builder contained  hard-coded DB credentials which revealed the  DB URL and username in the code. So I replaced them by passing them to the method as CLI args. Fixes #304


mir-am (approved):
LGTM! Thanks for the fix.

Issue #148, opened 2020-11-27
JavaCGOpal CLI does not generate call graph json files for dependencies.


cg122:
## Problem

JavaCGOpal CLI merging only generates files of raw call graph of 'artifact' and resolved call graph of 'artifact'. 

The call graphs of dependencies are generated but write to the same output file and finally overwritten by the call graph of 'artifact'.

Run command:

```
-a path/to/artifact -d path/to/deps -s -ma CHA -o path/to/output_file
```

The generated files are:  'output_file' and 'output_file_merged'

## Expected

All the raw and resolved call graph of dependencies are also write to separate files.

## Solution

As showed in a quick hack of Main.java of JavaCGOpal, only works for files as input. 

https://github.com/fasten-project/fasten/blob/sig/javacgopal-cli-merge/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/Main.java

Run command:
```
-a path/to/artifact -d path/to/deps -s -ma CHA -o path/to/output_directory
```
The generated files are: 'artifact.json', 'deps.json', 'artifact_merged.json', and 'dips_merged.json'

Issue #147, opened 2020-11-27
Support various format of input in JavaCGOpal CLI


cg122:
## Problem
Currently, JavaCGOpal CLI only supports maven coordinates or JAR(or Class) files as input for merging and the artifact and dependencies must be of the same format.

## Expected Behaviour

- More format of input, JavaCGOpal CLI supports JSON files of call graph as inputs.

For example, one can use `-m JSON` to indicate JSON call graph files as input.

- More options of `-m`, JavaCGOpal CLI supports mixed type (mode) of input.

The `-m` option requires artifact (provided with `-a` option) and dependencies (provided with `-d` option) in the same format (both coordinates or both JAR files). It would be necessary to have artifact and dependencies in the different format (modes).

For example, client code base is a JAR file, and dependencies are a list of maven coordinates, one can use `-m FILE-COORD` to indicate format of inputs for merging.

PR #406, opened 2022-02-09
Refactor opal Main

ashkboos:
This PR improves the CLI of OPAL and does some refactorings mostly on its Main file. It also considers requests in #147 and #148.


mir-am (approved):
LGTM!

PR #410, opened 2022-02-16
Generalize Rest-API to support C & Python

gdrosos:
## Description
This PR aims to generalize the REST-API plugin to support C & Python Knowledge Bases on the following ways:
* Modified the spring boot application properties & configuration to serve only one ecosystem per deployment.
* Removed `mvn` prefix from REST-API endpoints.
* Detected that some endpoints would require language specific handling, so we added an extra parameter (forge) to the application properties, based on which a language specific handling occurs at some methods.
* Moreover, modified some endpoints that previously worked only for java, to now support all ecosystems based on the forge variable. Those endpoints (previously broken but now functional) are the following:
`/package_version/{id}/rcg`
`packages/{pkg}/{pkg_ver}/modules/callables`
`packages/{pkg}/{pkg_ver}/rcg`
`packages/{pkg}/versions`
`/vulnerabilities/coordinates`
* Extended the Resolution API to support Dependency & Dependents Resolution for all languages. For the dependents resolution, we merged the DebianResolutionAPI and the PypiResolutionAPI methods into Resolution API. To support dependency resolution, we internally send GET requests to the dependency resolver microservice of either C or Python, whose URL is passed on the application through the application properties file. Now, those 3 resolution endpoints work for all languages:
`/packages/{pkg}/{pkg_ver}/resolve/dependencies`
`/packages/{pkg}/{pkg_ver}/resolve/dependents`
`/resolve_dependencies`

Now, all endpoints that either retrieve information from the Metadata Database or resolve dependents/dependencies are working for all languages.
## Motivation 
Since C & Python Databases contain basic & metadata information regarding each package version and its callables, it is only natural to properly extend the REST-API in order to share this knowledge.

## Testing
Tested all the aforementioned endpoints locally and they have the expected behaviour.


mir-am:
I will review this PR tomorrow.

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/PackageVersionApi.java
```diff
@@ -29,24 +29,46 @@
...
+                var groupId = coordinate[0];
+                var artifactId = coordinate[1];
+                var version = coordinate[2];
+                url = String.format("%smvn/%s/%s/%s_%s_%s.json", KnowledgeBaseConnector.rcgBaseUrl,
                 artifactId.charAt(0), artifactId, artifactId, groupId, version);
-        result = result.replace("\\/", "/");
-        return new ResponseEntity<>(result, HttpStatus.OK);
+                break;
+            }
+            case "pypi": {
```
MagielBruntink, line 53: There is a class with the supported forges by FASTEN, I suggest to use that instead of the hard-coded strings:
https://github.com/fasten-project/fasten/blob/develop/core/src/main/java/eu/fasten/core/data/Constants.java

analyzer/restapi-plugin/src/main/resources/application.properties
```diff
@@ -4,15 +4,13 @@ server.tomcat.connection-timeout=60000
...
-kb.c.url=${db.c.url}
-kb.maven.depgraph.path=${dg.maven.path}
-kb.pypi.depgraph.path=${dg.pypi.path}
-kb.debian.depgraph.path=${dg.debian.path}
+kb.url=${db.url}
+kb.depgraph.path=${dg.path}
 kb.graphdb.path=${gd.path}
 kb.vulnchains.path=${vc.path}
 lima.rcg.url=${rcg.url}
+dependency.resolver.address=${dep.api}
```
MagielBruntink, line 12: What is this variable supposed to contain?

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
```diff
@@ -883,18 +886,16 @@ public String getPackageVersions(String packageName, int offset, int limit) {
         Packages p = Packages.PACKAGES;
         PackageVersions pv = PackageVersions.PACKAGE_VERSIONS;
         ArtifactRepositories ar = ArtifactRepositories.ARTIFACT_REPOSITORIES;
-
         // Query
         var queryResult = context
                 .select(pv.ID, pv.PACKAGE_ID, pv.CG_GENERATOR, pv.VERSION, ar.REPOSITORY_BASE_URL, pv.ARCHITECTURE, pv.CREATED_AT)
                 .from(p)
                 .innerJoin(pv).on(p.ID.eq(pv.PACKAGE_ID))
-                .innerJoin(ar).on(pv.ARTIFACT_REPOSITORY_ID.eq(ar.ID))
```
mir-am, line 892: Why `innerJoin` is changed to `leftJoin` here?

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/PackageApi.java
```diff
@@ -144,24 +144,44 @@
...
-                return new ResponseEntity<>(result, HttpStatus.OK);
-            } else {
-                LazyIngestionProvider.ingestArtifactIfNecessary(packageName, version, artifactRepo, releaseDate);
-                return new ResponseEntity<>("Package version not found, but should be processed soon. Try again later", HttpStatus.CREATED);
+                break;
             }
-        } catch (IOException e) {
-            return new ResponseEntity<>(e.getMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
+            case "pypi": {
+                url = String.format("%s/%s/%s/%s/%s/cg.json", KnowledgeBaseConnector.rcgBaseUrl + KnowledgeBaseConnector.forge + "/" + KnowledgeBaseConnector.forge,
```
MagielBruntink, line 168: There is a bug here: KnowledgeBaseConnector.forge is included twice!

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/GraphResolverInitializer.java
```diff
@@ -0,0 +1,15 @@
+package eu.fasten.analyzer.restapiplugin;
+
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.client.RestTemplate;
+
+public class GraphResolverInitializer {
+
+    public static void sendRequestToInitGraphResolver() {
+        RestTemplate restTemplate = new RestTemplate();
+        String url = "http://localhost:8080/packages/package/0/resolve/dependents?transitive=false";
```
MagielBruntink, line 10: It's not a good idea to have this hardcoded URL here.

analyzer/restapi-plugin/src/main/resources/application.properties
MagielBruntink, line 12: The code seems to automatically prepend http:// to this value, which I don't think is a good idea.
In deployment you should be able to write: `--dep.api=http://fasten-pypi-resolver:5000` but now you can only do `--dep.api=fasten-pypi-resolver:5000`.

MagielBruntink:
Overall: fantastic work @gdrosos a big step forwards to having all languages/ecosystems supported on the same level!

I put some comments related to code. Other observations:

- If you try the resolution endpoints while the dependency graphs are not updated (need to restart REST API for that...) there is a generic error page without info on what is going on, eg. `http://localhost:9080/api/debian/packages/anna/1.71/resolve/dependents`
- `http://localhost:9080/api/debian/packages/anna/1.71/resolve/dependencies` generates a `{
error: "E:Unabletolocatepackageanna"
}` message that I think comes from the resolver microservice. Not sure how to fix that.

MagielBruntink:
I'm testing the new REST API with this branch of docker-compose: https://github.com/fasten-project/fasten-docker-deployment/tree/generalized-rest-api

analyzer/restapi-plugin/src/main/resources/application.properties
gdrosos: Hello, this variable contains the address of the dependency resolver microservices that have been deployed on monster: https://github.com/fasten-project/k8s-deployments/tree/master/fasten/resolvers
For the docker compose, a similar deployment with that of the stitching microservices (https://github.com/fasten-project/fasten-docker-deployment/pull/21) should take place in order to deploy the rest-api

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/PackageApi.java
gdrosos: Well, I intentionally included in the lima url  twice the forge, because for some reason, both on monster and on lima the rcg path includes two pypi directories. For example for the shap:0.34.0 revision the rcg url is the following: https://lima.ewi.tudelft.nl/pypi/pypi/callgraphs/s/shap/0.34.0/cg.json . (https://lima.ewi.tudelft.nl/pypi/callgraphs/s/shap/0.34.0/cg.json returns a 404 error)

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/GraphResolverInitializer.java
gdrosos: Previously, we sent this get request with the hard-coded url to init the grah maven resolver: ```http://localhost:8080/__INTERNAL__/packages/0/directedgraph``` , so to be honest I followed the same logic. The reasoning of this initialization is explained briefly on this [comment](https://github.com/fasten-project/fasten/pull/351/files#r762173190). If we want to hide this information, I would also suggest removing the hard-coded server port from this [line](https://github.com/fasten-project/fasten/blob/0af2d600b5da89fe4ecc77b707d86c9f0e1a6bb2/analyzer/restapi-plugin/src/main/resources/application.properties#L4).

analyzer/restapi-plugin/src/main/resources/application.properties
gdrosos: Agreed. Addressed on this commit: 49c9c91

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/PackageVersionApi.java
gdrosos: Thanks, changed this accordingly on faeb4ef

core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
gdrosos: Well, this method is used to retrieve all package versions of a specific package name. With inner join, it would return only package versions which have an artifact repository, which is a java  language specific attribute.  Therefore, with inner join, for C and Python the querry would not return any package version. To tackle this issue, changed the inner join to left join, so that to return also package versions that do not include an artifact repository attribute & table population.

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/PackageApi.java
MagielBruntink: Ugh, I see. Well it's easy to adjust this in the PyCG deployment to write to a /pypi/pypi path as well. Or the Delft guys @mir-am  could adjust the path on lima to make things look a bit nicer :-)

MagielBruntink: I think it's not nice to have such strong dependencies on file system arrangements in the code, though.

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/GraphResolverInitializer.java
MagielBruntink: I understand. Perhaps we could at least get the url out into the application.properties file?

MagielBruntink:
@gdrosos Issues are appearing because the directory structure no longer matches the (Java) packages. You removed `mvn` from the package path, but not from the directory.

analyzer/restapi-plugin/src/test/java/eu/fasten/analyzer/restapiplugin/mvn/api/PackageApiTest.java
```diff
@@ -46,7 +46,6 @@ void setUp() {
         service = new PackageApi();
         kbDao = Mockito.mock(MetadataDao.class);
         KnowledgeBaseConnector.kbDao = kbDao;
-        KnowledgeBaseConnector.forge = "mvn";
```
MagielBruntink: Hi @gdrosos This change is breaking your tests :-) You should add it back like:
`KnowledgeBaseConnector.forge = Constants.mvnForge;`

MagielBruntink: No problem, almost feels like pair programming :-)

analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/mvn/api/PackageApi.java
gdrosos: The lima url differs for each language: 
* https://lima.ewi.tudelft.nl/pypi/pypi/callgraphs/a/azureml-contrib-interpret/1.0.85/cg.json
* https://lima.ewi.tudelft.nl/debian/callgraphs/4/4digits/buster/1.1.4-1%2Bb1/amd64/file.json
* https://lima.ewi.tudelft.nl/mvn/z/zabbix-api/zabbix-api_io.github.hengyunabc_0.0.2.json
So I do not know if we can avoid this on the REST-API. Maybe on lima we could use a unified format for all languages

gdrosos:
> Overall: fantastic work @gdrosos a big step forwards to having all languages/ecosystems supported on the same level!
> 
> I put some comments related to code. Other observations:
> 
> * If you try the resolution endpoints while the dependency graphs are not updated (need to restart REST API for that...) there is a generic error page without info on what is going on, eg. `http://localhost:9080/api/debian/packages/anna/1.71/resolve/dependents`
> * `http://localhost:9080/api/debian/packages/anna/1.71/resolve/dependencies` generates a `{ error: "E:Unabletolocatepackageanna" }` message that I think comes from the resolver microservice. Not sure how to fix that.

Thanks for the detailed review. Regarding the second issue, I fixed it on 40cab69 by providing a consistent error message when a packsage/package version is not found. (Both C &Python microservices returned a json with an error key, so I just check if the returned string contains the ```"error"``` substring). Regarding the first issue, we should provide a more specific error message. Unfortunately, I have some obligations due to examinations, and I will try to address this and any other comment from the Weekend.

gdrosos:
> Overall: fantastic work @gdrosos a big step forwards to having all languages/ecosystems supported on the same level!
> 
> I put some comments related to code. Other observations:
> 
> * If you try the resolution endpoints while the dependency graphs are not updated (need to restart REST API for that...) there is a generic error page without info on what is going on, eg. `http://localhost:9080/api/debian/packages/anna/1.71/resolve/dependents`
> * `http://localhost:9080/api/debian/packages/anna/1.71/resolve/dependencies` generates a `{ error: "E:Unabletolocatepackageanna" }` message that I think comes from the resolver microservice. Not sure how to fix that.

Hello, regarding the error page of the dependents resolution, on 9de1533 I included a more informatve error message. Thanks both @MagielBruntink and @mir-am  for the immediate review and comments! Do not hesitate to point out anything else that needs to be fixed or refactored.

mir-am (approved):
What was the test plan for testing these changes?

gdrosos:
> What was the test plan for testing these changes?

Regarding the Unit Tests, for the endpoints that were modified to support C & Python I modified the tests to support the new methods, so that no test is failing.
Regarding testing those changes, I think that  @MagielBruntink created a branch in docker-compose to test this changes: https://github.com/fasten-project/fasten-docker-deployment/tree/generalized-rest-api But I do not know why, but it seems that right now this branch has been deleted.

MagielBruntink (approved):

MagielBruntink:
Nice!

PR #421, opened 2022-03-07
Avoid swallowing IOExceptions in the FASTEN server

mir-am:
## Description
This PR changes the FASTEN server to avoid swallowing IOExceptions when files cannot be written to the disk, e.g. permission errors or directory creation. Instead, it throws `RuntimeExceptions` to stop the plug-in.

## Motivation and context
Due to swallowing IOExceptions, the FASTEN server produced successfully processed records to `.err` Kafka topics. This fix now prevents committing Kafka offsets for IOExceptions that can be resolved.

## Testing
Tested with the DC.

### DIFF
--- a/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
+++ b/server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
@@ -18,23 +18,11 @@
 
 package eu.fasten.server.plugins.kafka;
 
-import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.net.InetAddress;
-import java.net.UnknownHostException;
-import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Properties;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


mir-am:
@ashkboos, I highly appreciate it if you could review this PR. It is a small change.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -414,13 +411,13 @@ private void emitMessage(KafkaProducer<String, String> producer, String topic, S
      * @param result message to write
      * @return Path to a newly written JSON file
      */
-    private String writeToFile(SingleRecord result) throws IOException, NullPointerException {
+    private String writeToFile(SingleRecord result) throws NullPointerException, IOException {
```
ashkboos, line 414: Do we still need to throw IOException in the method signature?
Also, I think throwing NPE in method signature here is a bit weird but I think your intention here was not fixing this so that's fine.

ashkboos (changes requested):

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
mir-am: Yes, `FileWriter` throws `IOException`. It is also caught in the `handle producing` method and wrapped in `RuntimeException` to stop the plug-in whenever `IOException` happens in the FASTEN server.

mir-am:
@ashkboos, do you approve this PR for merging?

ashkboos:
> @ashkboos, do you approve this PR for merging?

Yes, I don't have any other comments

PR #433, opened 2022-03-25
Fix the methods for finding vulnerable pacakge versions/callables in MetadataDao

mir-am:
## Description
Specifically, this PR makes changes to the `findVulnerablePackageVersions` and `findVulnerableCallables` methods by using the introduced vulnerability tables in #316.

## Motivation and context
After the merge of #316, the `vulnerabilities` field is no longer stored in the metadata field of the `package_versins` and `callables`. Therefore, it is necessary to adapt the mentioned methods for finding vulnerable package versions/callables.

## Testing
Tested with several vulnerable package versions and callables using the production metadata DB.

## Additional context
The method  `findVulnerablePackageVersions` needs this fix for the vulnerability chain finder [here](https://github.com/fasten-project/data-processing/pull/11).

### DIFF
--- a/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
+++ b/core/src/main/java/eu/fasten/core/data/metadatadb/MetadataDao.java
@@ -1602,21 +1602,26 @@ public class MetadataDao {
         PackageVersions pv = PackageVersions.PACKAGE_VERSIONS;
         Modules m = Modules.MODULES;
         Callables c = Callables.CALLABLES;
+        Vulnerabilities v = Vulnerabilities.VULNERABILITIES;
+        VulnerabilitiesXPackageVersions vxp = VulnerabilitiesXPackageVersions.VULNERABILITIES_X_PACKAGE_VERSIONS;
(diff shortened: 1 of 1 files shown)


MagielBruntink:
@mir-am It is not true that vulnerability metadata is no longer inserted into callables and package-versions. That still happens, but the inserted data is more limited than before.

MagielBruntink:
But it's still a good idea to use the vulnerability tables instead of the JSON metadata field. So this change LGTM. There are some build errors to be fixed, however.

mir-am:
> But it's still a good idea to use the vulnerability tables instead of the JSON metadata field. So this change LGTM. There are some build errors to be fixed, however.

Thanks for looking into the PR.
The build errors happened after 789cb6d17d2af52592c60df549b0660b62544ff3. It is unrelated to this PR. Maybe @proksch knows how to fix the build error.

proksch:
Sorry, I was not aware that the Jacoco version bump had an effect on the build, it was running locally and also on the Windows GitHub runner... the problem should be fixed now (see #435).

mir-am:
@MagielBruntink, We've fixed the build error. Can I proceed with merging the PR?

MagielBruntink:
Go ahead!

PR #434, opened 2022-03-26
Quickfix: Connect to RocksDB & enable Lazy Ingestion only when initializing the Java REST-API

gdrosos:
## Description
On this PR, we modify the REST-API by performing the GraphDB connection only on the Java REST-API Deployment. 
Also, since C & Python deployments do not include Lazy Ingestion, we avoid creating a corresponding Kafka Producer .

## Motivation and context
After trying to deploy C & Python REST-API on monster, I came up with a bug, as the REST-API failed to initilialize due to not being able to connect to the Graph Database. This PR fixes this issue.

### DIFF
--- a/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/KnowledgeBaseConnector.java
+++ b/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/KnowledgeBaseConnector.java
@@ -206,7 +206,6 @@ public class KnowledgeBaseConnector {
                     System.exit(1);
                 }
                 logger.info("...Graph database connection established successfully.");
-              
             }
         }
     }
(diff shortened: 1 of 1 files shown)


analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/KnowledgeBaseConnector.java
```diff
@@ -195,25 +196,33 @@ public void setInitDepGraphRequest() {
      */
     @PostConstruct
     public void connectToGraphDB() {
-        logger.info("Establishing connection to the Graph Database at " + graphdbPath + "...");
-        try {
-            graphDao = RocksDBConnector.createReadOnlyRocksDBAccessObject(graphdbPath);
-        } catch (RuntimeException e) {
-            logger.error("Couldn't connect to the Graph Database", e);
-            System.exit(1);
+        switch (KnowledgeBaseConnector.forge) {
```
mir-am, line 199: Why using a `switch` statement here? You could use an if statement.

gdrosos: I used the `switch` statement for consistency purposes, as all forge checks in the REST-API have been performed `switch` statement s, but I changed it to an `if` statement accordingly.

mir-am (approved):
LGTM!

PR #440, opened 2022-04-15
Debian license detector debugging

michelescarlato:
## Description
A fix for the Debian license detector.
## Motivation and context
For the outbound license detection, the detector scan three keywords in the package root directory: copyright, license and readme.
If one of the three keywords matches, but the license is null, the detector throws an exception and doesn't produce correctly the payload.

## Testing
Tested locally (through the DebianAPIJavaLib) and in DC.


michelescarlato:
I was using an old version of develop.
I am opening another PR (#441 ).

Issue #452, opened 2022-05-04
Kafka plugins crash when either a priority or a normal topics is missing


MagielBruntink:
## Describe the bug
The heartbeat call does not checked whether the plugin is actually subscribed to a priority or normal topic, and then this happens:
```
java.lang.IllegalStateException: Consumer is not subscribed to any topics or assigned any partitions
at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1222)
at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1210)
at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.sendHeartBeat(FastenKafkaPlugin.java:561)
at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.handleConsuming(FastenKafkaPlugin.java:257)
at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.run(FastenKafkaPlugin.java:167)
at eu.fasten.server.plugins.kafka.FastenKafkaPlugin.start(FastenKafkaPlugin.java:205)
at java.base/java.util.ArrayList.forEach(Unknown Source)
at eu.fasten.server.FastenServer.run(FastenServer.java:202)
at picocli.CommandLine.executeUserObject(CommandLine.java:1919)
at picocli.CommandLine.access$1100(CommandLine.java:145)
at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2159)
at picocli.CommandLine.execute(CommandLine.java:2058)
at eu.fasten.server.FastenServer.main(FastenServer.java:435)
```

## To Reproduce
1. Try to start any FASTEN Kafka (Java) plugin without a priority topic on DC
2. Observe log

MagielBruntink:
I'm working on a fix.

PR #453, opened 2022-05-04
Fix the heartbeat call crashing the server plugin: check if consumer is subscribed to the heartbeat topic(s).

MagielBruntink:
Fix for issue #452 by adding a check if the consumer is indeed subscribed to the topics it wants to do the heartbeat on.


mir-am:
@MagielBruntink
Sorry, I just saw the issue now.
Thanks for the fix.

MagielBruntink:
No worries. I am releasing the fix to 0.0.7. I thought I should go ahead because the issue is quite annoying in production and the fix is simple.

PR #459, opened 2022-05-21
Generate Java call graphs from local JARs in `OPALPlugin`

mir-am:
## Description
- Changes the `DataWriter` interface to `DataRW` so that the plugins can read/write from/to the disk given a base directory. 
- `OPALPlugin` first tries to generate a call graph from a local JAR file if available. If not, it downloads the JAR file from Maven central like before.

## Motivation and context
`POMAnalyzer` stores JAR files to the disk when resolving dependencies. Therefore, it is desirable to generate CGs from local JAR files and also reduce the network activity of `OPALPlugin`. 

## Testing
Tested with the DC.

### DIFF
--- a/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/OPALPlugin.java
+++ b/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/OPALPlugin.java
@@ -18,21 +18,6 @@
 
 package eu.fasten.analyzer.javacgopal;
 
-import static eu.fasten.analyzer.javacgopal.data.CGAlgorithm.CHA;
-import static eu.fasten.analyzer.javacgopal.data.CallPreservationStrategy.ONLY_STATIC_CALLSITES;
-import static eu.fasten.core.maven.utils.MavenUtilities.MAVEN_CENTRAL_REPO;
-import static java.lang.System.currentTimeMillis;
-
-import java.io.File;
-import java.util.Optional;
-
-import org.json.JSONObject;
-import org.pf4j.Extension;
-import org.pf4j.Plugin;
-import org.pf4j.PluginWrapper;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import eu.fasten.analyzer.javacgopal.data.OPALPartialCallGraphConstructor;
 import eu.fasten.core.data.Constants;
 import eu.fasten.core.data.JSONUtils;
@@ -42,6 +27,23 @@ import eu.fasten.core.data.opal.exceptions.EmptyCallGraphException;
 import eu.fasten.core.data.opal.exceptions.MissingArtifactException;
 import eu.fasten.core.data.opal.exceptions.OPALException;
 import eu.fasten.core.plugins.AbstractKafkaPlugin;
+import eu.fasten.core.plugins.DataRW;
+import org.json.JSONObject;
+import org.pf4j.Extension;
+import org.pf4j.Plugin;
+import org.pf4j.PluginWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+import static eu.fasten.analyzer.javacgopal.data.CGAlgorithm.CHA;
+import static eu.fasten.analyzer.javacgopal.data.CallPreservationStrategy.ONLY_STATIC_CALLSITES;
+import static eu.fasten.core.maven.utils.MavenUtilities.MAVEN_CENTRAL_REPO;
+import static java.lang.System.currentTimeMillis;
 
 public class OPALPlugin extends Plugin {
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/eu/fasten/core/data/opal/MavenCoordinate.java
```diff
@@ -156,9 +156,15 @@ public String toProductUrl() {
...
-    
+
     @Override
     public String toString() {
-    	return getCoordinate();
+        return getCoordinate();
     }
+
+    public String toPath() {
+        return Paths.get(groupID.replace('.', '/'), artifactID, versionConstraint,
```
proksch, line 166: This requires a unix file system. Please use the system specific separator constant.

core/src/main/java/eu/fasten/core/data/opal/MavenCoordinate.java
```diff
@@ -156,9 +156,15 @@ public String toProductUrl() {
         var repo = mavenRepos.get(0);
         return this.toURL(repo) + "/" + artifactID + "-" + versionConstraint + "." + packaging;
```
proksch, line 157: This has not been changed, but does not make sense. Which URL is this supposed to create?

analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/OPALPlugin.java
```diff
@@ -76,9 +79,18 @@ public void consume(String kafkaRecord, ProcessingLane l) {
                 // Generate CG and measure construction duration.
                 logger.info("[CG-GENERATION] [UNPROCESSED] [-1] [" + mavenCoordinate.getCoordinate() + "] [NONE] ");
                 long date = json.optLong("releaseDate", -1);
-				this.graph = OPALPartialCallGraphConstructor.createPartialJavaCG(mavenCoordinate,
-                        CHA, date, artifactRepository, ONLY_STATIC_CALLSITES);
-                long duration = currentTimeMillis() - startTime; 
+                try {
+                    this.graph = OPALPartialCallGraphConstructor.createPCGFromLocalJar(mavenCoordinate, date,
```
proksch, line 83: Do not use exception handling for programming logic! Concrete proposal: There is no need to have two separate methods, one `createPartialCG` method should be enough, which just takes a path to the jar to be analyzed. The only difference is where the path comes from: either `mavenCoord.getPath()` exists or the .jar is downloaded and a path to the downloaded file is provided. As the jar should *always* exist locally in our setup, I would be fine with just throwing an error when the file is missing.

analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/OPALPlugin.java
```diff
@@ -76,9 +79,18 @@ public void consume(String kafkaRecord, ProcessingLane l) {
...
+                            Paths.get(baseDir, mavenCoordinate.toPath()).toString(),
+                            CHA, ONLY_STATIC_CALLSITES);
+                } catch (FileNotFoundException e) {
+                    logger.warn(e.getMessage());
```
proksch, line 87: When logging, make sure that you preserve the stack trace.

As elaborated before, we can just remove this catch handling for now...

analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/OPALPartialCallGraphConstructor.java
```diff
@@ -127,12 +126,39 @@ public static PartialJavaCallGraph createPartialJavaCG(
...
 
+    /**
+     * Creates RevisionCallGraph from a local JAR given a Maven coordinate.
+     *
+     * @param coordinate maven coordinate of the revision to be processed
+     * @param timestamp  timestamp of the revision release
+     * @return RevisionCallGraph of the given coordinate.
+     * @throws FileNotFoundException
+     */
+    public static PartialJavaCallGraph createPCGFromLocalJar(final MavenCoordinate coordinate,
```
proksch, line 143: Please merge the two `createPCG` methods into one. There is NO need for a case distinction, both should do exactly the same...

core/src/main/java/eu/fasten/core/data/opal/MavenCoordinate.java
proksch: Stupid question, why not simply return a `Path`?

proksch (changes requested):
I think that the solution is sub-optimal... a case distinction that should be taken on a high level is dragged down through the various levels of the API. I have proposed a different strategy in the individual comments.

core/src/main/java/eu/fasten/core/data/opal/MavenCoordinate.java
mir-am: Because, when joining paths using `Paths.get()`, `String` is needed. I've changed the method to return `Path` now.

core/src/main/java/eu/fasten/core/data/opal/MavenCoordinate.java
mir-am: The method `toProductURL()` is not written by me. Based on its usage, I can tell it is used to download Maven artifacts like JARs.

core/src/main/java/eu/fasten/core/data/opal/MavenCoordinate.java
mir-am: Addressed in [3b12da6](https://github.com/fasten-project/fasten/pull/459/commits/3b12da65de924db0fa46c81047276b27866a104a)

mir-am:
@proksch, thanks for the code review and comments.
I have addressed your comments. This is not the best solution but it leverages our existing JARs in `.m2` and reduces the network traffic. Also, it downloads JARs, if not present in FS.

Issue #464, opened 2022-05-30
MavenResolverIO assumes dep_graph directory is present and crashes if it isn't


MagielBruntink:
I'm assuming this is related to recent changes to the REST API / Maven Resolver implementation there.

Repo: with DC:

- update FASTEN REST API image to 0.0.8 in .env
- clean `docker-volumes` directory
- start the pipeline including the REST API, eg `docker-compose --profile java up -d`
- observe log of `fasten-docker-deployment_fasten-restapi-java_1`

```
2022-05-30 13:31:44.030  INFO 1 --- [nio-8080-exec-1] e.f.c.maven.resolution.MavenResolverIO   : Saving poms to /mnt/fasten/dep_graph/poms.json ...
2022-05-30 13:31:44.035 ERROR 1 --- [nio-8080-exec-1] e.f.a.restapiplugin.api.ResolutionApi    : Error constructing dependency graph maven resolver

java.lang.RuntimeException: java.io.FileNotFoundException: /mnt/fasten/dep_graph/poms.json-tmp (No such file or directory)
at eu.fasten.core.maven.resolution.MavenResolverIO.saveToDisk(Unknown Source) ~[core-0.0.8.jar!/:na]
at eu.fasten.core.maven.resolution.MavenResolverIO.loadResolver(Unknown Source) ~[core-0.0.8.jar!/:na]
```

MagielBruntink:
It doesn't crash and seems to work if you create the `dep_graph` directory manually first.

mir-am:
I also experienced this issue. It can be solved by manually creating the `dep_graph` directory.

mir-am:
The issue is addressed in #467.

Issue #461, opened 2022-05-23
The REST API stops working if `PSQLException` happens.


mir-am:
On rare occasions, the Postgres DB connection may be disconnected and hence the REST API stops working due to `PSQLException`. For instance, this happens if the Postgres DB is restarted.
A fix is needed to kill the REST API's container in this case so that it'll be restarted with a new DB connection.

MagielBruntink:
Instead of killing & restarting, could we figure out why this exception / the disconnect is happening? It doesn't seem limited to restarting the postgres DB. 

If you want only kill & restart, the REST API should crash on the PSQLException, not log and proceed. Kubernetes will then restart the process.

Here is an example log:
```
2022-05-30 08:36:27.393 DEBUG 1 --- [nio-8080-exec-9] o.s.web.servlet.DispatcherServlet        : GET "/packages/com.nryanov.genkai:genkai-cats_2.13/0.6.0-RC1/deps", parameters={}
2022-05-30 08:36:27.393 DEBUG 1 --- [nio-8080-exec-9] s.w.s.m.m.a.RequestMappingHandlerMapping : Mapped to eu.fasten.analyzer.restapiplugin.api.DependencyApi#getPackageDependencies(String, String, int, int, String, Long)
2022-05-30 08:36:27.394 DEBUG 1 --- [nio-8080-exec-9] o.s.web.servlet.DispatcherServlet        : Failed to complete request: org.jooq.exception.DataAccessException: SQL [select "public"."packages"."id", "public"."packages"."package_name", "public"."packages"."forge", "public"."packages"."project_name", "public"."packages"."repository", "public"."packages"."created_at" from "public"."packages" join "public"."package_versions" on "public"."packages"."id" = "public"."package_versions"."package_id" where (true and lower("public"."packages"."package_name") = lower(?) and lower("public"."package_versions"."version") = lower(?))]; This connection has been closed.
[...]

PR #465, opened 2022-05-30
Stop/restart the Spring app (REST API) if `PSQLException` occurs

mir-am:
## Description
This PR adds a custom exception handling for the REST API when `PSQLException` occurs. For this exception, the exception handler stops the REST API app/container so that it will be restarted by K8s or DC to create a new PSQL DB connection.

## Motivation and context
As discussed in #461 and #464, the REST API gives an internal error (500) if its DB connection is closed.

## Testing
Tested with the DC.

### DIFF
--- a/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/ExceptionHandling.java
+++ b/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/ExceptionHandling.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2022 Delft University of Technology
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


MagielBruntink:
In the current state on monster, this will crash + restart the REST API all the time. That is maybe preferable to the current state in which the container keeps running but is actually pumping the log full of errors. With this PR I don't expect the overall problem to go away.

mir-am:
> In the current state on monster, this change will crash + restart the REST API all the time. That is maybe preferable to the current state in which the container keeps running but is actually pumping the log full of errors. With this PR I don't expect the overall problem to go away.

This is part of the solution that Sebastian (@proksch ) and I have devised today to solve the issue. Overall, our devised solution is as follows:
### Code level
- Catch `PSQLException` and stop the plugins/containers in the new loader and the FASTEN server including the REST API.
### Infrastructure level
- Tune the Postgres' configs like allocated memory and timeouts.
- Monitoring syslogs and Postgres' logs to see if any DB-related errors or warnings happen.
- Make `monster` unschedulable in K8s to make sure that the node won't be heavily loaded and hence disrupting the Postgres server.

MagielBruntink:
Can't k8s be configured to limit its resource usage on a node? Given that many analyses are still running I think not using monster's capabilities might be a set back.

mir-am:
> Can't k8s be configured to limit its resource usage on a node? Given that many analyses are still running I think not using monster's capabilities might be a set back.

We can use the [node-pressure evicition](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/) feature by K8s to terminate pods if the node's CPU or memory usage exceeds a certain threshold.

MagielBruntink (approved):

PR #466, opened 2022-05-31
Handle `PSQLException` in the FASTEN server and `MetadataDBExtension`

mir-am:
## Description
- The FASTEN server stops a plugin with a failed DB connection, that is, in the case where a `PSQLException` has occurred in the plugin.
- If `PSQLException` has happened in `MetadataDBExtension`, the plugin will be stopped and restarted by K8s or DC.

## Motivation and context
As discussed in #461, if a `PSQLException` happens in a plugin, it cannot proceed further due to a failed DB connection. Therefore, the plugin should be stopped and restarted by K8s/DC to make a new DB connection. 

## Testing
Tested with the DC.

### DIFF
--- a/analyzer/debian-license-feeder/src/main/java/eu/fasten/analyzer/debianlicensefeeder/DebianLicenseFeederPlugin.java
+++ b/analyzer/debian-license-feeder/src/main/java/eu/fasten/analyzer/debianlicensefeeder/DebianLicenseFeederPlugin.java
@@ -2,9 +2,6 @@ package eu.fasten.analyzer.debianlicensefeeder;
 
 import eu.fasten.core.data.Constants;
 import eu.fasten.core.data.metadatadb.MetadataDao;
-//This import should probably be different
-import eu.fasten.core.maven.data.Revision;
-import eu.fasten.core.plugins.AbstractKafkaPlugin;
 import eu.fasten.core.plugins.DBConnector;
 import eu.fasten.core.plugins.KafkaPlugin;
 import org.jooq.DSLContext;
@@ -17,8 +14,13 @@ import org.pf4j.PluginWrapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.sql.Timestamp;
-import java.util.*;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+//This import should probably be different
 
 public class DebianLicenseFeederPlugin extends Plugin {
 
@@ -82,7 +84,7 @@ public class DebianLicenseFeederPlugin extends Plugin {
 
             } catch (Exception e) { // Fasten error-handling guidelines
                 logger.error(e.getMessage(), e.getCause());
-                setPluginError(e);
+                throw e;
             }
         }
 
--- a/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
+++ b/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
@@ -174,16 +174,15 @@ public abstract class MetadataDBExtension implements KafkaPlugin, DBConnector {
                     }
                 });
             } catch (DataAccessException e) {
-                // The error codes starting with 57P0 are related to the DB connection issues.
-                // See https://www.postgresql.org/docs/current/errcodes-appendix.html
-                if (e.sqlState().contains("57P0")) {
-                    throw new UnrecoverableError("Could not connect to the Postgres DB and the plug-in should be stopped and restarted.",
+                transactionRestartCount++;
+                if (transactionRestartCount >= Constants.transactionRestartLimit) {
+                    throw new UnrecoverableError("Could not connect to or query the Postgres DB and the plug-in should be stopped and restarted.",
                             e.getCause());
                 }
             } catch (Exception expected) {
-
+                transactionRestartCount++;
             }
-            transactionRestartCount++;
+
         } while (restartTransaction && !processedRecord
                 && transactionRestartCount < Constants.transactionRestartLimit);
     }
--- a/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
+++ b/analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
@@ -18,14 +18,18 @@
 
 package eu.fasten.analyzer.metadataplugin;
 
-import eu.fasten.core.data.*;
+import eu.fasten.core.data.Constants;
+import eu.fasten.core.data.PartialCCallGraph;
+import eu.fasten.core.data.PartialCallGraph;
+import eu.fasten.core.data.PartialJavaCallGraph;
+import eu.fasten.core.data.PartialPythonCallGraph;
 import eu.fasten.core.data.callableindex.ExtendedGidGraph;
-import eu.fasten.core.data.callableindex.GidGraph;
 import eu.fasten.core.data.metadatadb.MetadataDao;
 import eu.fasten.core.data.metadatadb.codegen.enums.Access;
 import eu.fasten.core.data.metadatadb.codegen.enums.CallableType;
 import eu.fasten.core.data.metadatadb.codegen.tables.records.CallSitesRecord;
 import eu.fasten.core.data.metadatadb.codegen.tables.records.CallablesRecord;
+import eu.fasten.core.exceptions.UnrecoverableError;
 import eu.fasten.core.maven.utils.MavenUtilities;
 import eu.fasten.core.plugins.DBConnector;
 import eu.fasten.core.plugins.KafkaPlugin;
@@ -49,7 +53,11 @@ import java.io.IOException;
 import java.nio.file.Paths;
 import java.sql.BatchUpdateException;
 import java.sql.Timestamp;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 
 public abstract class MetadataDBExtension implements KafkaPlugin, DBConnector {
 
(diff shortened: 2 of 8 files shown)


analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
```diff
@@ -166,6 +174,14 @@ public void consume(String record) {
                     }
                 });
             } catch (Exception expected) {
+                if (expected instanceof DataAccessException) {
```
proksch, line 177: See below... better crate a second catch block.

analyzer/debian-license-feeder/src/main/java/eu/fasten/analyzer/debianlicensefeeder/DebianLicenseFeederPlugin.java
```diff
@@ -82,7 +84,7 @@ public void consume(String record) {
 
             } catch (Exception e) { // Fasten error-handling guidelines
                 logger.error(e.getMessage(), e.getCause());
-                setPluginError(e);
+                throw e;
```
proksch, line 87: I see that the exception is not consistently re-thrown across all plugins. What is the effect of that, doesn't this change the control flow in the application?

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
```diff
@@ -166,6 +174,14 @@ public void consume(String record) {
...
+                    // The error codes starting with 57P0 are related to the DB connection issues.
+                    // See https://www.postgresql.org/docs/current/errcodes-appendix.html
+                    if (((DataAccessException) expected).sqlState().contains("57P0")) {
```
proksch, line 180: Why do we need this weird state check? Can't we just catch `PSQLException` and fail without further checks?

proksch (commented):
I added several comments.

server/src/main/java/eu/fasten/server/plugins/kafka/FastenKafkaPlugin.java
```diff
@@ -354,6 +355,14 @@ public void processRecord(ConsumerRecord<String, String> record, ProcessingLane
             logger.error("Forced to stop the plug-in due to ", e);
             throw e;
         } catch (Exception e) {
+            if (e instanceof DataAccessException) {
```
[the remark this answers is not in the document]
mir-am: Addressed in [6948de1](https://github.com/fasten-project/fasten/pull/466/commits/6948de169066492fdb8a24447d2a0d96d4af31cc)

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
mir-am: Addressed in [6948de1](https://github.com/fasten-project/fasten/pull/466/commits/6948de169066492fdb8a24447d2a0d96d4af31cc)

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
mir-am: I used `if (expected.getCause() instanceof PSQLException)` and it didn't work oddly! So I have added a SQL state check. Also, note that `PSQLException` is a general DB exception and it may not necessarily be a DB connection error. Therefore, it would be better to check SQL states for more precise DB error handling.

mir-am:
> I added several comments.

@proksch
Thanks for reviewing the code and comments.
 I have addressed your comments and responded to your questions.

MagielBruntink:
Hi Amir, I tried restarting the REST API component for Java/MVN and it won't even come up anymore. I think it's really time to start trying things. What do you think?

mir-am:
> Hi Amir, I tried restarting the REST API component for Java/MVN and it won't even come up anymore. I think it's really time to start trying things. What do you think?

Hi Magiel (@MagielBruntink )
Would you please elaborate a bit more on this issue? Did you try this on `monster`?
Also, note that #465 is not running in production as a new release should be created.

MagielBruntink:
If I `kubectl rollout restart deployment/fasten-restapi-java -n fasten` on monster the Java/Mvn REST API container is restarted, but upon starting it immediately crashes.

mir-am:
> If I `kubectl rollout restart deployment/fasten-restapi-java -n fasten` on monster the Java/Mvn REST API container is restarted, but upon starting it immediately crashes.

I have created issue #471 as it's not related to this PR.

analyzer/debian-license-feeder/src/main/java/eu/fasten/analyzer/debianlicensefeeder/DebianLicenseFeederPlugin.java
proksch: At least for PomAnalyzer, I have just removed that retry mechanism to keep the handling consistent and simple. Throwing the unrecoverable error will also retry to connect to the DB.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
proksch: Ok, as it turns out, the thrown exception is a `DataAccessException` (jOOQ), not a `PSQLException`(Postgres). I really do not like this parsing of the SQL state, because as soon as a tiny detail changes, our handling will fail without us knowing. I would prefer to just catch *all* `DataAccessExceptions` and wrap them in an `UnresolvableError`.

proksch: I have just looked a bit into the details:
1) If it should indeed be required to distinguish exception states, please do not use `sqlState()` as this is vendor specific, but use `sqlStateClass()`, which is generic. On top of that, the latter can be conveniently checked through jOOQ's `SQLStateClass` enumeration, so we do not need to have magic strings like `"57P0"` in the code.
2) That being said, all `"57P0"` cases seem to cover serious database problems (e.g., DB shutdown) and not just transaction locks. As a result, I would advise to keep the simple solution for now that I have stated before (just wrap all `DataAccessExceptions` in `UnresolvableErrors`), because it is the quickest solution that guarantees data consistency. Once it is live, we should watch the logs closely, should there be indeed a locking problem, we can always try to find a solution that is less wasteful and does not restart the pod for every mistake.

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
```diff
@@ -105,6 +110,13 @@ public void consume(String kafkaMessage) {
                 catch(DataAccessException e) {
                     logger.info("Data access exception");
                     // Database connection error
+                    // The error codes starting with 57P0 are related to the DB connection issues.
+                    // See https://www.postgresql.org/docs/current/errcodes-appendix.html
+                    if (e.sqlState().contains("57P0")) {
```
proksch, line 115: See previous discussion.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
```diff
@@ -165,7 +173,15 @@ public void consume(String record) {
                         setOutputPath(callgraph);
                     }
                 });
+            } catch (DataAccessException e) {
+                // The error codes starting with 57P0 are related to the DB connection issues.
+                // See https://www.postgresql.org/docs/current/errcodes-appendix.html
+                if (e.sqlState().contains("57P0")) {
```
proksch, line 179: See previous discussion

analyzer/debian-license-feeder/src/main/java/eu/fasten/analyzer/debianlicensefeeder/DebianLicenseFeederPlugin.java
proksch: If I understand it correctly, the error handling (logging and setting error) is now entirely performed in the `FastenKafkaPlugin`. Why don't we just completely get rid of all these `try/catch` blocks in all plugins? These methods only re-throw `e` and do not seem to add any value to the error handling...

proksch (changes requested):

analyzer/quality-analyzer/src/main/java/eu/fasten/analyzer/qualityanalyzer/QualityAnalyzerPlugin.java
mir-am: Addressed in [b9dbc27](https://github.com/fasten-project/fasten/pull/466/commits/b9dbc27c6b91895e7a28b43d985207bf883e22b7).

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
mir-am: Addressed in [b9dbc27](https://github.com/fasten-project/fasten/pull/466/commits/b9dbc27c6b91895e7a28b43d985207bf883e22b7).

mir-am:
@proksch
Thanks for the discussion and further comments.
I have addressed them in my recent commits.

analyzer/metadata-plugin/src/main/java/eu/fasten/analyzer/metadataplugin/MetadataDBExtension.java
proksch: Seems to be resolved in latest changes.

analyzer/debian-license-feeder/src/main/java/eu/fasten/analyzer/debianlicensefeeder/DebianLicenseFeederPlugin.java
proksch: Seems to be resolved in the last commits

proksch (approved):

PR #467, opened 2022-06-01
Make `MavenResolverIO` to create the dependency graph directory if it doesn't exist 

mir-am:
## Description
This is a tiny PR to make `MavenResolverIO` create the dependency graph directory if it doesn't exist.

## Motivation and context
As reported in #464, the `MavenResolverIO` assumes that the dep. graph directory already exists. However, this is not always the case, especially when running the FASTEN DC.

## Testing
Tested within the IDE when running the REST API.

### DIFF
--- a/core/src/main/java/eu/fasten/core/maven/resolution/MavenResolverIO.java
+++ b/core/src/main/java/eu/fasten/core/maven/resolution/MavenResolverIO.java
@@ -15,7 +15,16 @@
  */
 package eu.fasten.core.maven.resolution;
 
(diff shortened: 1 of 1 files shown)


MagielBruntink (approved):

MagielBruntink:
I just had this issue re-appear on a Linux machine. The REST API crashes during start-up if the `poms.json` file is not there, or if that file is empty, it appears. Putting a `poms.json` file with just `[]` in it resolves the issue.

Issue #471, opened 2022-06-07
Java REST API crashes due to the out of memory exception


mir-am:
## Describe the bug
The REST API is completely unresponsive when it's just launched. A simple query like `http://localhost/packages` does not work.

## Logs
```
2022-06-07 10:08:21.062 DEBUG 1 --- [nio-8080-exec-1] o.s.web.servlet.DispatcherServlet        : Failed to complete request: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'resolutionApi' defined in URL [jar:file:/app/rest-api.jar!/BOOT-INF/classes!/eu/fasten/analyzer/restapiplugin/api/ResolutionApi.class]: Instantiation of bean failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [eu.fasten.analyzer.restapiplugin.api.ResolutionApi]: Constructor threw exception; nested exception is java.lang.OutOfMemoryError: Java heap space
2022-06-07 10:08:21.068 ERROR 1 --- [nio-8080-exec-1] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'resolutionApi' defined in URL [jar:file:/app/rest-api.jar!/BOOT-INF/classes!/eu/fasten/analyzer/restapiplugin/api/ResolutionApi.class]: Instantiation of bean failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [eu.fasten.analyzer.restapiplugin.api.ResolutionApi]: Constructor threw exception; nested exception is java.lang.OutOfMemoryError: Java heap space] with root cause

java.lang.OutOfMemoryError: Java heap space

```

## Environment (please complete the following information):
 - OS: Ubuntu 18.04
 - Java version: 11

proksch:
Right now, it tries to instantiate the MavenResolver, which likely has issues with the new DB size. How much memory does the REST API have at the moment? Until we factor out the resolution to the micro-service, the REST API should have at least 32GB, if not more by now.

mir-am:
To give an update, with 64 GB of the heap size, the REST API still crashes. See: 
```
$ kubectl top pod -n fasten fasten-restapi-java-5bcb74b56d-tsbv8
NAME                                   CPU(cores)   MEMORY(bytes)   
[...]

PR #472, opened 2022-06-08
Use lazy fetch in `MavenResolverIO` when reading POMs from the DB

mir-am:
## Description
Specifically, this PR makes the following changes:
- Use JOOQ's `lazyFetch()` in `readFromDB()` to avoid loading lots of table rows in a list and causing `OOM`.
- Set JDBC `AutoCommit` to false when creating a PG connection in the REST API. This is needed as PG doesn't like non-zero fetch size and `autocommit == false` together. See [this](https://www.jooq.org/doc/latest/manual/sql-execution/fetching/lazy-fetching/#fetch-sizes) for more info.

## Motivation and context
As described in #471, the Java REST API currently crashes due to `OOM`, which is caused by reading all package versions' metadata into memory. This should be done lazily, i.e., one by one.

## Testing
Tested with the DC and within the IDE. 
Also, with this fix, in production, the Java REST API is now responsive and the `OOM` exception does not occur.

### DIFF
--- a/analyzer/restapi-plugin/Dockerfile
+++ b/analyzer/restapi-plugin/Dockerfile
@@ -3,6 +3,6 @@ FROM openjdk:11-jre-slim
 WORKDIR /app
 COPY ./target/*.jar ./rest-api.jar
 
-ENTRYPOINT java $JAVA_OPTS -jar /app/rest-api.jar $0 $@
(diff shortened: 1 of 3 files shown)


proksch (approved):
After the clarifications in the call, this change looks good to me

PR #481, opened 2022-07-27
Fix documentation

amottier:
Updating the deployment schema to reflect latest updates on k8s and Docker Compose deployment.


gdrosos:
Hello @amottier, thanks for the documentation. As a side note: the Debian & PyPI resolver are connected to the REST-API by the endpoints `{package}{version}/resolve/dependencies` and therefore are exposed online. Their address is given as an argument in the REST-API deployment, e.g. here https://github.com/fasten-project/k8s-deployments/blob/2e64df7526cd9886470bd2126d93491994d56598/fasten/rest-api/python-restapi-deployment.yaml#L19  the `--dependency.resolver.address`

PR #486, opened 2022-08-30
Catch `ArrayIndexOutOfBoundsException` when deserializing the `Exclusion` object

mir-am:
## Description
This is a tiny PR to ignore `ArrayIndexOutOfBoundsException` when deserializing `Exclusion` objects that contain malformed Maven coordinates.

## Motivation and context
This issue currently makes the dependency graph resolver crash all the time and hence hinders its progress

## Testing
Not Yet.


proksch:
I have revised the proposed solution a bit:

1) I have created an implementation that is not based on exceptions, but that simply checks the array size.
2) I have added a regression test case to reflect the problem and prevent it from happening again.

Next, I will fix the `PomAnalyzer` to not create such broken JSON serializations in the first place.

proksch:
I think we broke the review system, as I cannot select you as a reviewer for your own PR. I have approved the change from my side, please review it and just merge it, if you agree.

mir-am (commented):
The new changes are okay.

PR #489, opened 2022-09-07
Trim trailing and leading whitespace when getting upper/lower bound versions

mir-am:
## Description
This is a small fix to the `getUpperBound` and `getLowerBound` methods in `VersionConstraint` to trim trailing and leading whitespace.  

## Motivation and context
Whitespace in version strings might cause issues, for example, when downloading Jars or doing comparisons.

## Testing
Added a unit test.

### DIFF
--- a/core/src/main/java/eu/fasten/core/maven/data/VersionConstraint.java
+++ b/core/src/main/java/eu/fasten/core/maven/data/VersionConstraint.java
@@ -15,13 +15,13 @@
  */
 package eu.fasten.core.maven.data;
 
-import static eu.fasten.core.utils.Asserts.assertNotNull;
-import static eu.fasten.core.utils.Asserts.assertTrue;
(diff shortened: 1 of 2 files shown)


proksch:
The problem with this class is that it is central for the (de-) centralization of `Pom` objects... I would prefer not to change it without deeper analysis of the effect of the changes on performance.

I also do not see the use case, how would you create a download URL for a version range? There is no way you can figure out all versions in between, for that, you need the resolution service, which does return a `ResolvedRevision` instead...

proksch:
If white spaces remain indeed an issue, this should be addressed in the extractor that "fills" the VersionConstraint.... we could simply remove all spaces (or any whitespace for that matter), they do not carry any meaning in a version.

mir-am:
I agree with the points above. Also, I am using `RestMavenResolver` in `core` to get dependencies with a resolved version. So, there seems to be no need to merge this PR.
I am going to close this PR.

PR #490, opened 2022-09-15
Optimization to building OPAL Class Hierarchies

mir-am:
## Description
This is a small PR, which creates a new method `asURIHierarchyParallel()` to create Fasten URIs in parallel. The speed up is 2 times, especially for large graphs.

## Motivation and context
When creating (very large) partial call graphs, building FURI hierarchies can be very time-consuming if it's done in single-thread mode.

## Testing
Unit tests pass for the OPAL plugin.

### DIFF
--- a/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/analysis/OPALClassHierarchy.java
+++ b/analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/analysis/OPALClassHierarchy.java
@@ -18,16 +18,12 @@
 
 package eu.fasten.analyzer.javacgopal.data.analysis;
 
-import java.lang.reflect.InvocationTargetException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.EnumMap;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
+import eu.fasten.analyzer.javacgopal.data.CallPreservationStrategy;
+import eu.fasten.core.data.FastenURI;
+import eu.fasten.core.data.JavaGraph;
+import eu.fasten.core.data.JavaScope;
+import eu.fasten.core.data.JavaType;
+import it.unimi.dsi.fastutil.ints.IntIntPair;
 import org.opalj.br.ClassHierarchy;
 import org.opalj.br.DeclaredMethod;
 import org.opalj.br.Method;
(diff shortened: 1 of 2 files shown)


ashkboos (approved):

analyzer/javacg-opal/src/main/java/eu/fasten/analyzer/javacgopal/data/analysis/OPALClassHierarchy.java
```diff
@@ -234,17 +262,18 @@ public Map<Object, Object> getInternalMetadata(final Map<List<Integer>, Map<Obje
...
+     * @param targets      list of targets
+     * @param resultGraph  already existing PartialJavaCallGraph
      * @param callSiteOnly
      */
-    public void appendGraph(final Object source,
-                            final Iterator<Tuple2<Object, Iterator<DeclaredMethod>>> targets,
-                            final Stmt<DUVar<ValueInformation>>[] stmts,
-                            final JavaGraph resultGraph, List<Integer> incompeletes,
-                            final Set<Integer> visitedPCs, CallPreservationStrategy callSiteOnly) {
+    public synchronized void appendGraph(final Object source,
```
ashkboos, line 272: This synchronized method is not the most efficient way of parallelizing this part. I suggest that we make the data structure thread-safe. Data structures that we need to change would be OpalClassHierarchy and JavaGraph. Perhaps some ConcurrentMap AtomicInt would do the job. However, this is a bit complicated we may need to debug it together if needed.

mir-am: I agree. The most efficient solution should also make the data structures related to CGs thread-safe. For now, to pass tests, I've added `synchronized` to `appendGraph()`. 
It would be nice to work together and make `appendGraph()` more efficient for multi-threading.

mir-am:
@ashkboos, thanks a lot, Mehdi, for the contribution and help with the optimization of OPAL data structures.

mir-am:
@ashkboos 
What is the verdict on this PR? Shall we merge it into `develop`?

ashkboos:
> @ashkboos What is the verdict on this PR? Shall we merge it into `develop`?

To me it's fine. Unless Sebastian also wants to take a look.

mir-am:
I would like to request a review from @proksch as he recently refactored `OPALPartialCallGraphConstructor`.

MagielBruntink:
What is the status of this PR? Do we still want to merge it?

PR #491, opened 2022-09-23
Extend Maven Resolution

proksch:
Working with the Maven Resolution, it became obvious that minor extensions are needed. This PR adds options to...

- limit how deep into the dependency tree the dependency/dependents resolution step down/up
- limit the number of returned values (and abort remaining search)
- removes the requested coordinate from the list of returned dependencies

Once this PR is accepted, the options still need to be integrated into the resolution micro service that is being developed in the `data-processing` repository.


core/src/main/java/eu/fasten/core/maven/resolution/MavenDependentsResolver.java
```diff
@@ -49,7 +48,7 @@ public Set<ResolvedRevision> resolve(String gav, ResolverConfig config) {
         }
 
         var dependents = new HashSet<ResolvedRevision>();
-        resolve(pom, config, dependents, new HashSet<>(), false, COMPILE);
+        resolve(pom, config, dependents, new HashSet<>(), false, COMPILE, 1);
```
mir-am, line 51: What is this hard-coded number, `1` ? It's depth, I guess. Maybe it would be better to introduce a constant, like `DIRECT`

core/src/main/java/eu/fasten/core/maven/resolution/ResolverConfig.java
```diff
@@ -48,11 +48,34 @@ public ResolverConfig at(long timestamp) {
         return this;
     }
 
-    public ResolverConfig depth(ResolverDepth depth) {
+    public ResolverConfig includeTransitiveDeps() {
+        this.depth = Integer.MAX_VALUE;
+        return this;
+    }
+
+    public ResolverConfig excludeTransitiveDeps() {
```
mir-am, line 56: If the depth is 1, would it be better to name it `includeDirectDeps`?

mir-am (approved):
Thanks for the PR.
Overall, LGTM.
I have some minor comments.

core/src/main/java/eu/fasten/core/maven/resolution/ResolverConfig.java
proksch: That would be misleading as it is always the case... direct dependencies are also included when you resolve the transitive ones. The only difference in configuration is the extent at which transitive dependencies are included, so I have named the three relevant methods accordingly.

core/src/main/java/eu/fasten/core/maven/resolution/MavenDependentsResolver.java
proksch: This is the depth, yes. You can see this in the name of the method parameter. Given that the depth is constantly increased in the recursive calls, how would I use a constant here? On a side note, I would like to replace the recursive solution with a work queue anyways (which eliminates the need for this parameter), but did not find the time to do this in this refactoring...

PR #497, opened 2022-11-19
Exclude derived fields from Pom serialization

proksch:
The `Pom` class contain some pre-computed fields to improve performance of the Maven resolution. Erroneously, these fields are included in the serialization right now, which does not make sense as they get re-computed on de-serialization anyways. This PR removes the fields from serialization.


mir-am (approved):
LGTM!

PR #498, opened 2022-11-20
Several improvements to performance and code quality of the REST API

proksch:
During my investigation of the slow response time of the package version endpoint, I made three improvements in the REST API:

1) The `LazyIngestionProvider` was a static monster. I have made the first step towards dependency injection and made it a regular class. We should consider making it a `@Component` moving forward, so Spring can take care of the lifecycle management and the injection. An advantage of the current refactoring is also an improved testability of all dependent classes.
2) For unclear reasons, our config has limited the REST API to 20 threads, while even the default is 200 threads already. To prevent a potential bottleneck, I have removed this artificial limit.
3) I have introduced a minor performance improvement in the package version endpoint. Yet un-ingested packages result in a short-circuit and the logic will avoid an expensive, joined call to the database.

None of these improvements has solved the major performance issue (it looks like we were just missing an index), but this PR makes some small contributions to performance and code quality.


mir-am (approved):
LGTM!

PR #502, opened 2022-12-19
Fix Broken PyPI Ingestion Mechanism

### LINKED ISSUE #488: PyPI query used for ingestion no longer contains releases info (breaking pypi-filter)
The REST API uses the following URL in the lazy ingestion code, for example for namesss/0.1.0: https://pypi.org/pypi/namesss/0.1.0/json

However, the JSON returned by the PyPI API is now too limited; it no longer contains releases information. The PyPI Filter then logs: `Could not retrieve packaging info`

gdrosos:
## Description
This Pull Request aims to fix the issue described on #488 :

- Precisely, it enables the PyPI ingestion mechanism to check the existence of a PyPI coordinate and consequently consume packaging information throughout the use of the https://pypi.org/pypi/package_name/json endpoint instead of the https://pypi.org/pypi/package_name/package_version/json. We use the package-only endpoint because it is the only endpoint containing the json field with all the release information of the package. Consuming this information is essential, because further in the pipeline it enables PyPI Filter to ingest all the missing versions of a corresponding package.

- Moreover, in order to check whether a specific package version exists on PyPI, we search  whether the specific version of the package exists in  the `releases` key of the json response of the pypi.org package endpoint. 

## Motivation and context
Fixes #488

## Testing
Tested in a local deployment on monster, and the release information could be succesfully consumed by the kafka topic.

### DIFF
--- a/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/LazyIngestionProvider.java
+++ b/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/LazyIngestionProvider.java
@@ -117,12 +117,12 @@ public class LazyIngestionProvider {
      * @return whether it was necessary to ingest artifact
      */
     public boolean ingestPypiArtifactIfNecessary(String packageName, String version) {
-        var query = "https://pypi.org/pypi/" + packageName + "/" + version +"/json";
+        var query = "https://pypi.org/pypi/" + packageName + "/json";
         String result;
         try {
             result = MavenUtilities.sendGetRequest(query);
         } catch (IllegalStateException ex) {
-            throw new IllegalArgumentException("PyPI package " + packageName + ":" + version
+            throw new IllegalArgumentException("PyPI package " + packageName
                     + " could not be found. Make sure the PyPI coordinate is correct");
         }
         if (!hasArtifactBeenIngested(packageName, version)) {
--- a/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/LazyIngestionProvider.java
+++ b/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/LazyIngestionProvider.java
@@ -126,10 +126,19 @@ public class LazyIngestionProvider {
                     + " could not be found. Make sure the PyPI coordinate is correct");
         }
         if (!hasArtifactBeenIngested(packageName, version)) {
-            JSONObject json = new JSONObject(result);
+
+            //Check if the specific package version exists on PyPI
+            JsonObject releases = JsonParser.parseString(result).getAsJsonObject()
+                                    .get("releases").getAsJsonObject();
+            if (!releases.has(version)) {
+                throw new IllegalArgumentException("PyPI package version " + packageName + ":" + version
+                + " could not be found. Make sure the PyPI coordinate is correct");
+            }
+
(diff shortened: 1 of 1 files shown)


MagielBruntink (commented):
Hi @gdrosos thanks for the fix! Overall the ingestion works; I tested it locally in the docker-compose setup on several coordinates and they are injected fine.

One weird thing I observe is that after requesting a package-version that is not yet in the DB, you first get the expected message: `"Package version not found, but should be processed soon. Try again later."`

if you then refresh right away, you get `"Package version not found"`

then refresh again a bit later and the package-version appears. The `"Package version not found"` message might be a bit confusing as the analysis is actually on-going. For Maven I think the API repeats the `"Package version not found, but should be processed soon. Try again later."` (with HTTP 201 - CREATED), which I feel is somewhat better.

MagielBruntink:
The package/version endpoint Op 20 dec. 2022 om 13:34 heeft Giorgos Drosos ***@***.***> het volgende geschreven:﻿

Hi @gdrosos thanks for the fix! Overall the ingestion works; I tested it locally in the docker-compose setup on several coordinates and they are injected fine.
One weird thing I observe is that after requesting a package-version that is not yet in the DB, you first get the expected message: "Package version not found, but should be processed soon. Try again later."
if you then refresh right away, you get "Package version not found"
then refresh again a bit later and the package-version appears. The "Package version not found" message might be a bit confusing as the analysis is actually on-going. For Maven I think the API repeats the "Package version not found, but should be processed soon. Try again later." (with HTTP 201 - CREATED), which I feel is somewhat better.

Thanks! Which specific PyPI endpoint reproduces this error?

gdrosos:
> The package/version endpoint.

The fact that this issue occurs only in PyPI is really weird, because the package api code is generic for all languages. I see that there have been some recent changes in the architecture of the Rest-API modules, and I can see that now we perform 2 checks. https://github.com/fasten-project/fasten/blob/fd2f97f325b2be92476e1b755aa77d019098c714/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/api/PackageApi.java#L74-L82
 When the artifact is already ingested but not yet in the db, the package version sql querry returns null leading to the  `packageVersionNotFound` Exception.
The reason that this happens only to PyPI might be that in PyPI we directly update the ingestedArtifacts table.
https://github.com/fasten-project/fasten/blob/fd2f97f325b2be92476e1b755aa77d019098c714/analyzer/restapi-plugin/src/main/java/eu/fasten/analyzer/restapiplugin/LazyIngestionProvider.java#L134

Based on the current  `ingestMvnArtifactIfNecessary` implementation, I think recieving the message _"Package version not found, but should be processed soon. Try again later."_  twice indicates that the maven coordinate is being ingested twice in the pipeline, which should be avoided.

MagielBruntink:
I think you are right - in the Maven case multiple requests for ingestion are put into the pipeline in the Maven case. However, the pom-analyzer prevents that work is done multiple times, so the problem does not propagate further.

Overall I think the message "Package version not found." while actually ingestion is in progress is a bit confusing. Especially right after the API told the user that they should come back later. Perhaps we should just improve the message in that particular case? Any thoughts on that?

Also: @proksch since you were working a bit on this part of the code, do you have an opinion on it?

MagielBruntink (approved):
Ok, looks good, let's get this merged. Let's also address the logic of `hasArtifactBeenIngested` in a separate PR.

PR #503, opened 2023-03-02
Escape < > characters in FastenPythonURI creation.

MagielBruntink:
## Description
PyCG generates namespaces that can include `<` and `>` to denote lambdas. The creation of FastenPythonURIs does not take that into account and hence fails. See issue https://github.com/fasten-project/pypi-tools/issues/15.

This PR escapes the `<` and `>` characters so that URI creation succeeds.

## Testing
Unit tests succeed locally.

### DIFF
--- a/core/src/main/java/eu/fasten/core/data/FastenPythonURI.java
+++ b/core/src/main/java/eu/fasten/core/data/FastenPythonURI.java
@@ -18,6 +18,8 @@
 
 package eu.fasten.core.data;
 
+import com.google.common.net.UrlEscapers;
+
 import java.net.URI;
 
 /** A class representing a Fasten URI for the Python language; it has to be considered experimental
@@ -29,7 +31,7 @@ public class FastenPythonURI extends FastenURI {
     protected final boolean isFunction;
 
     public FastenPythonURI(final String s) {
-        this(URI.create(s));
+        this(URI.create(UrlEscapers.urlFragmentEscaper().escape(s)));
     }
 
     public FastenPythonURI(final URI uri) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/eu/fasten/core/data/FastenPythonURI.java
```diff
@@ -29,7 +31,7 @@ public class FastenPythonURI extends FastenURI {
     protected final boolean isFunction;
 
     public FastenPythonURI(final String s) {
-        this(URI.create(s));
+        this(URI.create(UrlEscapers.urlFragmentEscaper().escape(s)));
```
gdrosos, line 34: From the [documentation](https://guava.dev/releases/19.0/api/docs/com/google/common/net/UrlEscapers.html) of the `urlFragmentEscaper()` I saw that it converts most of the special characters (including `< >`)  into one or more bytes using UTF-8 encoding. If the URI somehow includes a special character which is not allowed to exist in the URI, this conversion will not break the processing of the specific package version. Will this use case create issues further down the pipeline or e.g. in the analysies using the URI or it won't be a problem?

gdrosos (approved):
LGTM, very clean sollution!
I had only one concern regarding the `urlFragmentEscaper()` method,  but if we are sure that it won't create issues further down  in the PyPI pipeline/ PyPI analysies then we can merge.

core/src/main/java/eu/fasten/core/data/FastenPythonURI.java
MagielBruntink: I don't it should be an issue. The FASTEN URIs will be correctly encoded versions of the namespaces that PYCG produces into its call graphs. If downstream issues will emerge we should fix them there, but I doubt whether it will occur.

gdrosos: Ok, I agree, let’s merge this!

